Issue #4, opened 2023-02-04
ENS Support


QuintonC:
When a wallet is connected via the `connect-wallet` package, we should look up if the wallet also has an associated ENS and display it in the connected wallet state instead of the truncated address.

### Concerns and challenges
- This should **only** be the desired functionality when:
  - Either an Alchemy or Infura provider are present
  - Why? 
    - Minimize the risk of rate limiting for developers and the storefronts which these components may be rendered on.
- How to handle cross chain support (could defer this until #3 is complete)

```[tasklist]

- [x] Initial ENS Support - Shopify/blockchain-components#8
- [ ] Loading ENS name component [Show spinner while loading ENS in ConnectButton]
- [ ] Decision: ENS Avatar support?
- [ ] Decision: Should the address pill allow folks to select either their ENS or their address?
- [ ] Decision: Should the address pill allow folks to select either their ENS or their address?
```

PR #8, opened 2023-02-08
:sparkles: Adds support for ENS resolution

QuintonC:
⚠️ Part of: https://github.com/Shopify/blockchain-components/issues/4

Adds support for ENS resolution.

> ❗️ Note
> This makes use of the provider functions which can potentially lead to rate limited providers. For the time being, we will allow the use of ENS resolution without non-public providers being present. However, we will output a console warning when the developers are utilizing only a public provider.

This expands on our tests and test utilities as well as providing ens forward and reverse resolution. That means we use the address provided at connection to look up an ENS name (if present), and we then utilize forward resolution to double check that the ENS name provided points back to the address we originally looked up for.

## Demonstration

<details><summary>Video demonstration</summary>
<p>

https://user-images.githubusercontent.com/4250423/219260209-2a4ea2a9-f450-49af-b38b-6903da68659f.mov

</p>
</details> 

## How can this be tophatted?

I will provide a CodeSandbox sample after a `/snapit` command has been run.

[CodeSandbox Demo](https://fxj3fg-5173.preview.csb.app/)

## 🎩 Checklist

- [x] Tested on mobile
- [x] Includes unit tests

### DIFF
--- a/packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
+++ b/packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
@@ -19,7 +19,9 @@ import {Popover} from './Popover';
 import {CaretIcon, ConnectedButton, Wrapper} from './style';
 
 export const ConnectButton = () => {
-  const {connectedWallets} = useAppSelector((state) => state.wallet);
+  const {activeWallet, connectedWallets} = useAppSelector(
+    (state) => state.wallet,
+  );
   const {openModal} = useModal();
   const [popoverVisible, setPopoverVisible] = useState(false);
   const {t} = useTranslation('ConnectButton');
@@ -42,12 +44,12 @@ export const ConnectButton = () => {
   }, [escPress, popoverVisible, togglePopover]);
 
   const handleClick = useCallback(() => {
-    if (!connectedWallets.length) {
+    if (!connectedWallets.length || !activeWallet) {
       openModal();
     }
-  }, [connectedWallets.length, openModal]);
+  }, [activeWallet, connectedWallets.length, openModal]);
 
-  if (!connectedWallets.length) {
+  if (!activeWallet) {
     return (
       <Button
         aria-label={t('buttonText')}
@@ -60,7 +62,7 @@ export const ConnectButton = () => {
     );
   }
 
-  const {address, connectorId} = connectedWallets[0];
+  const {address, connectorId, displayName} = activeWallet;
 
   return (
     <Wrapper id="connectWalletConnectedButtonWrapper" ref={ref}>
@@ -72,7 +74,7 @@ export const ConnectButton = () => {
       >
         <ConnectorIcon id={connectorId} size="Xs" />
         <Text as="span" variant="bodyLg">
-          {formatWalletAddress(address)}
+          {displayName || formatWalletAddress(address)}
         </Text>
         <CaretIcon>{CaretDown}</CaretIcon>
       </ConnectedButton>
--- a/packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
+++ b/packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
@@ -19,9 +19,7 @@ import {Popover} from './Popover';
 import {CaretIcon, ConnectedButton, Wrapper} from './style';
 
 export const ConnectButton = () => {
-  const {activeWallet, connectedWallets} = useAppSelector(
-    (state) => state.wallet,
-  );
+  const {activeWallet} = useAppSelector((state) => state.wallet);
   const {openModal} = useModal();
   const [popoverVisible, setPopoverVisible] = useState(false);
   const {t} = useTranslation('ConnectButton');
@@ -44,10 +42,10 @@ export const ConnectButton = () => {
   }, [escPress, popoverVisible, togglePopover]);
 
   const handleClick = useCallback(() => {
-    if (!connectedWallets.length || !activeWallet) {
+    if (!activeWallet) {
       openModal();
     }
-  }, [activeWallet, connectedWallets.length, openModal]);
+  }, [activeWallet, openModal]);
 
   if (!activeWallet) {
     return (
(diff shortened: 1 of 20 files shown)


packages/connect-wallet/src/utils/provider.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (isFallbackProvider) {
+    return provider.providerConfigs.every(
+      ({provider: currentProvider}: {provider: EthereumProviderType}) =>
+        'connection' in currentProvider &&
+        currentProvider.connection.url === defaultUrl,
+    );
+  } else {
+    return 'connection' in provider && provider.connection.url === defaultUrl;
+  }
+};
```
QuintonC, line 36: These were extracted because at one point during a pairing session, @Soleone mentioned that they might be of use outside of the context of this feature implementation.

packages/connect-wallet/src/test/providers.ts
```diff
@@ -0,0 +1,74 @@
+import {providers} from 'ethers';
+
+/**
+ * Copied from @wagmi/chains.
+ *
+ * This is only temporary because Jest doesn't support ESM.
+ * Perhaps we can consider Vitest?
+ */
```
QuintonC, line 8: Left a similar note about migrating to Vitest in another comment on this pull request.

packages/connect-wallet/src/test/fixtures/signature.ts
```diff
@@ -15,15 +16,15 @@ export const MOCK_MESSAGE_DATA = {
 export const MOCK_MESSAGE = new SiweMessage(MOCK_MESSAGE_DATA);
 
 export const VALID_SIGNATURE_RESPONSE: SignatureResponse = {
-  address: MOCK_MESSAGE.address,
+  address: MOCK_MESSAGE.address as Address,
```
QuintonC, line 19: address is a string in `MOCK_MESSAGE` and defining the type in the message didn't carry the type down to this construct for some reason.

Probably not the cleanest way to handle this so feel free to provide an alternative approach 👍

packages/connect-wallet/src/providers/ConnectWalletProvider/types.ts
```diff
@@ -0,0 +1,64 @@
+import {Chain} from '@wagmi/core';
```
QuintonC, line 1: This is all relocated from `src/types/provider.ts`

Soleone:
I haven't looked at the code, but I tested it, it works for normal domain and subdomain, but I noticed a delay, also visible in your video, where the address first shows up and then gets replaced by the ENS name.

Could we do one of the following:

a) Show a loading indicator next to the address.
b) Eager fetch the ENS during the signing step similarly to how it was done in the delegate cash?

packages/connect-wallet/src/slices/walletSlice/attributeOrder.test.ts
```diff
@@ -12,6 +12,11 @@ jest.mock('./gateContextClient', () => {
 });
 
 describe('attributeOrder', () => {
+  // Removes the orderAttribution console.error from the test output.
+  beforeAll(() => jest.spyOn(console, 'error').mockImplementation(() => {}));
```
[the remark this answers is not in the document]
QuintonC: > I think instead of silencing the error, we should investigate how to actually remove it.

The error is expected (it's part of the order attribution feature set with logging errors when it catches one).

I'm only ignoring it in our test suites because it makes it look like tests have broken.

packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -19,7 +19,9 @@ import {Popover} from './Popover';
 import {CaretIcon, ConnectedButton, Wrapper} from './style';
 
 export const ConnectButton = () => {
-  const {connectedWallets} = useAppSelector((state) => state.wallet);
+  const {activeWallet, connectedWallets} = useAppSelector(
```
jamiely, line 22: Can you briefly explain the difference between `activeWallet` and `connectedWallets`? Intuitively, I would think that the activeWallet is equivalent to the last `connectedWallet`, but some of the code here makes me think otherwise. Is it just that 
Can you briefly explain the difference between `activeWallet` and `connectedWallets`? Intuitively, I would think that the activeWallet is equivalent to the last `connectedWallet`, but some of the code here makes me think otherwise. Is it just that the `activeWallet` may be `disconnected`, and we don't set the `activeWallet` to the last connected wallet after the current one? How should I think about this?

packages/connect-wallet/src/utils/provider.ts
```diff
@@ -0,0 +1,36 @@
...
+export const isDefaultProvider = ({chain, provider}: ProviderCheckProps) => {
+  const isFallbackProvider = isFallbackProviderType(provider);
+
+  const defaultUrl = chain.rpcUrls.default.http[0];
+
+  if (isFallbackProvider) {
+    return provider.providerConfigs.every(
+      ({provider: currentProvider}: {provider: EthereumProviderType}) =>
+        'connection' in currentProvider &&
+        currentProvider.connection.url === defaultUrl,
```
jamiely, line 31: Good as-is. I'd consider extracting this lambda for reuse on line 34.

jamiely (approved):
Looks good to me. I don't have an ENS so I'm not sure how I could tophat. 

Are you still planning on adding tests for `useMiddleware`? No rush, just wondering.

packages/connect-wallet/src/utils/provider.ts
QuintonC: Addressed in the most recent commit 👍

QuintonC:
> Looks good to me. I don't have an ENS so I'm not sure how I could tophat.

Sent you a message about this 👀 

> Are you still planning on adding tests for `useMiddleware`? No rush, just wondering.

Yep! I'd like to do that at some point. I started adding some test infra (only a file or two) which would allow me to add those tests and then stashed it because the PR was growing too much. I'm also open to someone else picking that work up if they'd like!

caropinzonsilva (approved):
Looks good!

It was kind of hard to review since it is addressing too many changes. For the sake of feedback, I think this could have been a good way of splitting it:
- Moving of types to other file
- Changes in the attributeOrder test
- fetchEns slice + tests without it being used
- Addition of fetchEns into the middleware
- Changes in the provider

PR #13, opened 2023-02-09
Implement Delegate cash support

Soleone:

This way end users can unlock tokengates or claim rewards by signing messages from a wallet without any expensive assets in it while keeping their assets in secure vault addresses that don't require any further messages signed from.

## Demonstration

TODO

    <details>
      <summary>Summary of your gif(s)</summary>
      Description of what the gif shows
    </details>

## How can this be tophatted?

TODO

## 🎩 Checklist

- [ ] Tested on mobile


packages/shared/src/icons/Link.tsx
```diff
@@ -0,0 +1,9 @@
+export const Link = (
```
QuintonC, line 1: Is this used anywhere? I couldn't find any usage of it 🤔

jamiely (commented):
Do we plan to merge this? If not, can we close?

PR #18, opened 2023-02-14
[connect-wallet] Add `Playground` story for Connect Wallet package

caropinzonsilva:

This PR adds a Playground to Storybook for the `Connect Wallet` package. The idea is to use the tools Storybook provides to enhance the development tools. 

## Demonstration
16-23-2kvc9-n09dk

## How can this be tophatted?
Can be 🎩 in https://639b1f308693132693d9b82c-cafbriamhq.chromatic.com/?path=/story/connect-wallet--playground&args=wallets:Ethereum

## 🎩 Checklist

- [x] Tested on mobile
- [x] Updated relevant documentation for the changes (if necessary)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v1
 
       - name: Install dependencies
-        run: yarn install
+        run: yarn install && yarn build
 
       - name: Publish to Chromatic
         run: npx chromatic --project-token=${{ secrets.CHROMATIC_PROJECT_TOKEN }}
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -6,6 +6,10 @@ name: 'Chromatic'
 # Event for the workflow
 on: push
 
+env:
+  # 7 GiB by default on GitHub, setting this to 16 GiB
+  NODE_OPTIONS: --max-old-space-size=16384
+
 # List of jobs
 jobs:
   chromatic-deployment:
--- a/packages/connect-wallet/src/components/Playground/Playground.stories.tsx
+++ b/packages/connect-wallet/src/components/Playground/Playground.stories.tsx
@@ -1,0 +1,48 @@
+import {WagmiConfig, configureChains, createClient} from 'wagmi';
+import {mainnet} from 'wagmi/chains';
+import {publicProvider} from 'wagmi/providers/public';
+
+import {ConnectWalletProvider} from '../../providers/ConnectWalletProvider';
+import {getDefaultConnectors} from '../../connectors/getDefaultConnectors';
+import {ConnectButton} from '..';
+
+const PlaygroundStory = {
+  title: 'Connect Wallet',
+  component: WagmiConfig,
+  argTypes: {
+    wallets: {
+      control: {type: 'select'},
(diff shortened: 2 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -14,7 +18,7 @@ jobs:
       - uses: actions/checkout@v1
 
       - name: Install dependencies
-        run: yarn install
+        run: yarn install && yarn build
```
caropinzonsilva, line 21: We had to add the build step. Otherwise, we were getting the error
>[commonjs--resolver] Failed to resolve entry for package "@shopify/gate-context-client". 

https://github.com/Shopify/blockchain-components/actions/runs/4176925673/jobs/7233845999#step:4:86

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {getDefaultConnectors} from '../../connectors/getDefaultConnectors';
+import {ConnectButton} from '..';
+
+const PlaygroundStory = {
+  title: 'Connect Wallet',
+  component: WagmiConfig,
+  argTypes: {
+    wallets: {
+      control: {type: 'select'},
+      options: ['Ethereum'],
```
caropinzonsilva, line 15: For now, for this select, we only have 1 option. But when we add more chains support, we can add them here.

.github/workflows/chromatic.yml
QuintonC: Should this be moved to a different job in the workflow? It looks like it's under `Install dependencies`, would it be beneficial to have a dedicated build step?

.github/workflows/chromatic.yml
```diff
@@ -6,6 +6,10 @@ name: 'Chromatic'
 # Event for the workflow
 on: push
 
+env:
+  # 7 GiB by default on GitHub, setting this to 16 GiB
+  NODE_OPTIONS: --max-old-space-size=16384
```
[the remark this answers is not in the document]
QuintonC: Hmm... interesting. Good find! 👍 

I wonder if this is solved in the most recent updates with Storybook. In another Repo I'm using their v7 (which is pre-release), and it seems to build much faster than what I've experienced before 🤔 

Maybe we can throw upgrade / test v7 of Storybook into the backlog.

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
QuintonC: Hmm... maybe we can consider having a couple of options here in a follow-up PR.

I'm thinking something that allows folks to select which `Platform` (Ethereum or Solana) and then which `Chain` they want to use. It might make it easier to test things like Polygon or Goerli support as well if we can add that :)

QuintonC (approved):
Looks good to me, left some comments to consider for follow-up tasks 👍

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
caropinzonsilva: Yeah, we should definitely add more wallets later whenever we have that support. 

I'm following the designs we have for here btw:
17-09-250ll-dl7c3

PR #19, opened 2023-02-14
[chore] :recycle: Refactor onConnect listener creation to a util function

QuintonC:

Refactors and simplifies the `onConnect` listener logic. This allows for easier middleware creation for `onConnect` events by removing the possibility to missed actions (if the actions for `onConnect` logic are to ever change) and DRYs up the creation of listeners.

The inspiration behind this was not only to DRY up the code a bit, but also because in #4 I am going to re-use this logic (it could likely be reused in #13 as well) for looking up ENS names when a wallet connects.

Note: I was going to add tests for `useMiddleware.ts` as well, however I noticed that we need to create some form of app context in order to render hooks with the state provider. I'd prefer for that to be done the right way, and I believe that doing so will make this PR larger than it should be.

## How can this be tophatted?

Primarily green CI (I've added tests for the functionality). However, I will also run `/snapit` and link a CodeSandbox with the snapshot version for testing.

[Test CodeSandbox](https://codesandbox.io/p/sandbox/shopify-connect-wallet-pr-19-hrseiz)

## 🎩 Checklist

- [x] Tested on mobile
- [x] Includes unit tests (see note in the context for the PR)

### DIFF
--- a/packages/connect-wallet/src/hooks/useConnectWallet/useConnectWalletCallbacks.ts
+++ b/packages/connect-wallet/src/hooks/useConnectWallet/useConnectWalletCallbacks.ts
@@ -1,58 +1,24 @@
-import {isAnyOf} from '@reduxjs/toolkit';
-import {useContext, useEffect} from 'react';
+import {useEffect} from 'react';
 
-import {ConnectWalletContext} from '../../providers/ConnectWalletProvider';
-import {
-  addWallet,
-  removeWallet,
-  setActiveWallet,
-  validatePendingWallet,
-} from '../../slices/walletSlice';
+import {buildOnConnectMiddleware} from '../../middleware/onConnectMiddleware';
+import {removeWallet} from '../../slices/walletSlice';
 import {addListener} from '../../store/listenerMiddleware';
-import {Wallet} from '../../types/wallet';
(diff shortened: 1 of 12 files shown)


QuintonC:
You can test this PR using [this CodeSandbox](https://codesandbox.io/p/sandbox/shopify-connect-wallet-pr-19-hrseiz)

packages/connect-wallet/src/hooks/useConnectWallet/useConnectWalletCallbacks.ts
```diff
@@ -21,37 +15,11 @@ export const useConnectWalletCallbacks = (props?: useConnectWalletProps) => {
...
-        },
-      }),
-    );
+    const listener = buildOnConnectMiddleware(({wallet}) => {
+      onConnect?.(wallet);
+    });
 
-    return unsubscribeToOnConnectListener;
+    return dispatch(listener);
   }, [dispatch, onConnect, requireSignature]);
```
jamiely, line 23: is `requireSignature` still a dependency?

jamiely (approved):
Great cleanup, thanks! I tophatted only on MetaMask Desktop

PR #27, opened 2023-02-22
:sparkles: Upgrades Storybook to v7 alpha version

QuintonC:
⚠️ Fixes: 

V7 uses only Vite as a builder and doesn't fallback to using Webpack 4 under the hood. This should create faster build times for Chromatic and also prevent any future issues with Webpack + Node version support.

This is just a sample of what would change with the migration to version 7 which will have breaking changes with the construction of stories (as you can see by the number of changes here).

It might be beneficial to get ahead of the curve, but it's also a little bit of a risk given this is an alpha release.

Note that I had to remove the theme provider addon because it does not work with v7 yet. However, I've added a custom addon for this feature which can be seen in the toolbar of Storybook (url shared below).

## 🕹️ Demonstration

[Preview Storybook](https://639b1f308693132693d9b82c-futaivomat.chromatic.com/)

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A


.github/workflows/chromatic.yml
```diff
@@ -25,7 +25,11 @@ jobs:
           cache: 'yarn'
 
       - name: Install dependencies
-        run: yarn install && yarn build
+        run: yarn install
+
+      # This omits the build of Playground
+      - name: Build packages
+        run: yarn build:packages
```
QuintonC, line 32: Updating to only build the packages instead of also building the playground saves us a bit of time during this CI step.

| | Workflow Duration |
| - | - |
| Before | 3m 13s |
| After | 2m 15s |

Frame 4

jamiely (approved):
LGTM, will this correct some of the dependency issues we see involving storybook?

caropinzonsilva (approved):
Looks great, thanks for making this change! I 🎩 using Chromatic

Issue #16, opened 2023-02-10
`Uncaught TypeError: storage.getItem is not a function` with create-react-app and polyfills


ryanberckmans:
I'm trying out connect-wallet with vanilla setup and am receiving `Uncaught TypeError: storage.getItem is not a function` from redux-persist using create-react-app and polyfills for `Buffer` and `process`.

I'm using https://github.com/niksy/node-stdlib-browser for the polyfills.

Is this error related to a missing polyfill?

Do you have an example of connect-wallet in create-react-app?

QuintonC:
Hey @ryanberckmans 👋 

I'm not certain what the cause of this error might be. My assumption might be similar to yours in that it is possibly a missing polyfill.

Would you be able to provide a sample app / repo that we can use for testing? If you could use a service like [CodeSandbox](https://codesandbox.io/) that would be great :) 

Not to dismiss the usage of Create React App, I do want to help resolve the issue, but I might recommend that if you're not set on using Create React App, it might be worthwhile to consider using more modern tooling such as [Vite](https://vitejs.dev/). 

Here is a [CodeSandbox](https://codesandbox.io/p/sandbox/shopify-connect-wallet-vite-example-9w7wv6) that uses Vite if you're interested in what that might look like.

ryanberckmans:
Hey 👋 . I think you're right that I'd benefit from switching to something like Vite. I've been meaning to make the switch.

My create-react-app is fairly ordinary, except that I'm using a [fork](https://github.com/facebook/create-react-app/compare/main...ryanberckmans:create-react-app:main) of create-react-app for polyfills and tailwind support

```
  "react-scripts": "https://gitpkg.now.sh/ryanberckmans/create-react-app/packages/react-scripts?08746634d19a73958ba9b764b3cda418e1fce151",
```
I'll switch to Vite asap, and then try connect-wallet again. I'm looking forward to it, the library looks quite competitive, thanks for making it.

PR #29, opened 2023-02-24
:bug: Fix exports to ensure shared types are being bundled

QuintonC:
⚠️ Fixes: 
🔗 Related: https://github.com/Shopify/blockchain-components/issues/16

There are some issues with the way the packages are bundled and the references for the packages.

- The npm.js package shows `shared` as a dependency.
  - To resolve this, we link to shared via paths in `tsconfig` and modify the eslint rules for imports to accept `shared` imports and group them as external modules.
- Prior to this PR, importing `Theme` from either of the packages (`connect-wallet` or `tokengate`) would actually result in an unknown type and have no declarations inside of it. This made overwriting and extending themes a mystery, and also resulted in `theme` being an unknown prop on the `ConnectWalletProvider` and `Tokengate` components.
- Addresses an import issue where `import {verifyMessage} from 'ethers/lib/utils';` would cause an error for Webpack bundlers as it imported from a path module. This has been updated to `import {utils} from 'ethers';` and `verifyMessage` calls are invoked via `utils.verifyMessage`
- Addresses an import issue where `import storage from 'redux-persist/lib/storage';` would cause an error for Webpack bundlers as it imported from a path module. Since `redux-persist` doesn't export storage directly from the package's root, I've copied the functionality from that package into our package.

... More details to come.

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests

### DIFF
--- a/packages/connect-wallet/package.json
+++ b/packages/connect-wallet/package.json
@@ -24,12 +24,6 @@
   ],
   "types": "./dist/index.d.ts",
   "main": "./dist/index.js",
-  "exports": {
-    ".": {
-      "types": "./dist/index.d.ts",
-      "default": "./dist/index.js"
-    }
-  },
   "scripts": {
     "build": "tsup",
     "dev": "DEV=true tsup --watch",
@@ -66,7 +60,6 @@
     "@types/styled-components": "^5.1.26",
     "@types/ua-parser-js": "^0.7.36",
     "@wagmi/core": "0.8.10",
-    "shared": "*",
     "tsconfig": "*",
     "tsup": "^6.5.0"
   }
--- a/packages/connect-wallet/package.json
+++ b/packages/connect-wallet/package.json
@@ -59,7 +59,6 @@
     "@types/redux-logger": "^3.0.9",
     "@types/styled-components": "^5.1.26",
     "@types/ua-parser-js": "^0.7.36",
-    "@wagmi/core": "0.8.10",
     "tsconfig": "*",
     "tsup": "^6.5.0"
   }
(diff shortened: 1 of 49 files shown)


packages/connect-wallet/package.json
```diff
@@ -65,8 +59,6 @@
     "@types/redux-logger": "^3.0.9",
     "@types/styled-components": "^5.1.26",
     "@types/ua-parser-js": "^0.7.36",
-    "@wagmi/core": "0.8.10",
```
QuintonC, line 68: `@wagmi/core` usage for Chains results in mixed Chain types and results in typecheck errors.

packages/connect-wallet/src/components/ConnectButton/style.ts
```diff
@@ -1,5 +1,4 @@
-import {breakpoints, ButtonWrapper} from 'shared';
-import styled from 'styled-components';
+import {breakpoints, ButtonWrapper, styled} from 'shared';
```
QuintonC, line 1: styled is imported from shared instead of styled-components in an attempt to resolve issues with Webpack bundlers which do not properly support ESM packages.

The issue here is that Webpack looks for styled.default (which doesn't exist on the default export for styled), which the styled export from shared exports.

packages/connect-wallet/src/components/Modal/style.ts
```diff
@@ -118,7 +118,9 @@ export const Sheet = styled.div`
   }
 `;
 
-export const SheetContent = styled.div<{rowGap?: CSSProperties['rowGap']}>`
+export const SheetContent = styled.div<{
+  rowGap?: CSSProperties['rowGap'];
+}>`
```
QuintonC, line 123: Not really certain why this needed to change, but I was getting ESLint violations 🤔

PR #31, opened 2023-02-27
:bug: Addresses an issue with exported types from shared

QuintonC:
⚠️ Fixes: 

`Dawn`, `Default`, and `Theme` were exported without any information since `packages/shared` is not a published package. 

This leads to issues when attempting to pass a theme value to not only `Tokengate`, but also `ConnectWalletProvider`. Additionally, this makes it impossible to understand the type definition of `Theme` as well as results in errors when attempting to override some values in any of the provided themes.

## 🕹️ Demonstration

| Package Name | Before | After |
| - | - | - |
| `@shopify/connect-wallet` |  image |  image |
| `@shopify/tokengate` | image | image |

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/.changeset/silver-cougars-check.md
+++ b/.changeset/silver-cougars-check.md
@@ -1,0 +1,6 @@
+---
+'@shopify/connect-wallet': patch
+'@shopify/tokengate': patch
+---
+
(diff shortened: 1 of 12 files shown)


caropinzonsilva (approved):
Looks good!

PR #34, opened 2023-02-27
:bug: Resolves issue with redux-persist path imports

QuintonC:
⚠️ Related: https://github.com/Shopify/blockchain-components/issues/16
🔨 Related PR: https://github.com/Shopify/blockchain-components/pull/33

> 💡 This is similar to #33, but addresses issues for a different package.

Resolves issue for webpack bundlers which cannot resolve path imports for redux-persist by internalizing the storage object which is used to construct the local storage type for persisted data.

The storage file is similarly constructed to the redux-persist package, but has been simplified slightly as we only need one storage object type (which is local storage). Additionally, it includes changes to the storage construction to ensure the file aligns with our ESLint config.

| Before | After |
| - | - |
|  image | Coming soon |

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests

### DIFF
--- a/.changeset/purple-drinks-double.md
+++ b/.changeset/purple-drinks-double.md
@@ -1,0 +1,5 @@
+---
+'@shopify/connect-wallet': patch
+---
+
+Resolves an issue with redux-persist storage import using path imports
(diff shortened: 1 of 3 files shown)


jamiely (approved):
:+1:

PR #37, opened 2023-02-28
:package: Patch Redux + use RTK alpha for full ESM support

QuintonC:

Addresses an issue with Redux Toolkit and ESM support. When attempting to use the package in Next.js an error would be present stating addListener was not exported. This was primarily due to the lack of support for ESM usage in Redux Toolkit, which [version 2](https://github.com/reduxjs/redux-toolkit/pull/3089) aims to resolve.

This makes use of an alpha package and a patch in order to remove an `EmptyObject` type from the Redux `CombinedState` type.

After chatting with some of the team from Redux, it appears the _only_ "downside" to removing `EmptyObject` from the `CombinedState` type is that we must ensure that our `initialState` passed to `combineReducers` has all keys defined (which we already did anyway).

## 🕹️ Demonstration

| State (w/ Demo) | Image |
| - | - |
| [Before](https://codesandbox.io/p/sandbox/charming-cookies-5bgjz7?file=%2FREADME.md) | SCR-20230302-n9k |
| [After](https://codesandbox.io/p/sandbox/cocky-jennings-2ttvfj?file=%2Fpackage.json&selection=%5B%7B%22endColumn%22%3A62%2C%22endLineNumber%22%3A13%2C%22startColumn%22%3A62%2C%22startLineNumber%22%3A13%7D%5D) | SCR-20230302-ng5 |

There are demo links on the demonstration :) 

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A


jamiely (approved):
We paired on this and it seems fine, but how will we maintain this? Does it require documentation? Should there be some issue to look into this again a couple months from now?

Also, the codesandbox links open and I understand they are for next, but should we have some place to tophat our regular installation (non-next)?

caropinzonsilva (approved):

PR #38, opened 2023-03-01
Feature/refactor build connectors

Borahm:
⚠️ Fixes: 

...

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/connect-wallet/src/hooks/useDefaultConnectors.ts
```diff
@@ -1,24 +0,0 @@
-import {useContext} from 'react';
-
-import {getDefaultConnectors} from '../connectors/getDefaultConnectors';
-import {ConnectWalletContext} from '../providers/ConnectWalletProvider';
-import {Connector, ConnectorInstance} from '../types/connector';
-
-export function useDefaultConnectors() {
```
jamiely, line 7: What are the thoughts on whether removing this requires a major update? Although it is not exported from the package index, it is exported from this file.

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
```diff
@@ -12,20 +13,39 @@ import {ProviderProps} from './types';
...
   const contextValue: ConnectWalletProviderValue = useMemo(() => {
+    let contextualConnectors = connectors;
+
+    if (!contextualConnectors.length) {
+      const {connectors: defaultConnectors} = buildConnectors({
+        appName: '',
+        chains,
+        includeDefaults: true,
+      });
+      contextualConnectors = defaultConnectors;
```
jamiely, line 32: It seems like this might be similar to what `getDefaultConnectors` does? If so, should it be used here? Also, should `getDefaultConnectors` be refactored to use `buildConnectors`?

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
```diff
@@ -25,18 +24,18 @@ const Template = ({wallets}: {wallets: 'Ethereum' | 'Solana'}) => {
     publicProvider(),
   ]);
 
-  const {connectors} = getDefaultConnectors({chains});
+  const {connectors, wagmiConnectors} = buildConnectors({appName: '', chains});
```
jamiely, line 27: It is a little confusing for me seeing `appName` blank. Should we make it optional? It seems only required for coinbase. Or maybe we can just improve the documentation.

packages/connect-wallet/src/connectors/buildConnectors.ts
```diff
@@ -0,0 +1,124 @@
...
+    MetaMask({chains}),
+    Coinbase({appName, chains}),
+    Rainbow({chains}),
+    LedgerLive({chains}),
+    WalletConnect({chains}),
+  ];
+
+  defaultAvailableConnectors.forEach(
+    ({createConnector, id, ...provided}: ConnectorInstance) => {
+      const createdConnector = createConnector();
```
caropinzonsilva, line 83: This can be moved to after the next two `ifs` since it is not used until then. This would improve code readability and  performance since we would only define it when we are going to use it.

packages/connect-wallet/src/connectors/buildConnectors.ts
```diff
@@ -0,0 +1,124 @@
...
+         * Context: MetaMask on mobile, Rainbow, etc. will use WalletConnect bridge
+         * and need to use a WC connector. These are all going to use the same WC
+         * connector from Wagmi. But, WalletConnect itself will use two different
+         * WC connectors. One for support with WalletConnect's modal and one without.
+         */
+
+        // When we begin adding more connectors (e.g. Ledger, Rainbow, etc.) we should
+        // take a look at this code again to make sure we're seeing the expected connectors.
+        // We should always have TWO walletConnect connectors (one with a qr code and one without).
+        if (connectors.some((item) => item.id === 'walletConnect')) {
```
caropinzonsilva, line 109: I think instead of having nested ifs, it would be better if we define this variable after the `isWalletConnect` with a descriptive name and then do the logic in the same if:
```javascript
const isWalletConnect = createdConnector.id === 'walletConnect';
const descriptiveName =  connectors.some((item) => item.id === 'walletConnect');
if (isWalletConnect && descriptiveName) {
  return;
}
```

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
caropinzonsilva: Good question. From [this change](https://github.com/Shopify/blockchain-components/pull/38/files#diff-755ee8ba3e3b0d00dc4e1e8713581e6ad441f004581ca7b525153ff4c7f75952L15) I thought `getDefaultConnectors` would get deleted.

packages/connect-wallet/src/hooks/useDefaultConnectors.ts
QuintonC: IMO this shouldn't require a major update.

This was only used internally to access the default connectors which are the same list of connectors returned via `getDefaultConnectors`.

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
QuintonC: I think we want to deprecate `getDefaultConnectors`, or instead just have `getDefaultConnectors` call `buildConnectors` internally.

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
QuintonC: Good point... I think we might have to overload it since it is required for Coinbase 🤔

QuintonC (dismissed):
I'm converting this PR to a draft and requesting changes. The "request changes" is only to block this from accidentally shipping to main when it's not quite ready.

1. This PR is based off of #32 and that PR is being merged into a feature branch, [`feature/customizable-wallet-list`](https://github.com/Shopify/blockchain-components/tree/feature/customizable-wallet-list).
    - It might be best to wait until that PR is merged into the feature branch before opening the PR for review.
2.  This PR should also be merging to the feature branch mentioned above rather than `main`

caropinzonsilva:
@Borahm @QuintonC I based the PR in the other branch and rebased it so that it only includes the new changes. We might need to do a rebase if @lesliexin force pushes any changes to that other PR, but this will unblock this PR from getting reviewed.

lesliexin:
@caropinzonsilva @Borahm @QuintonC Just force pushed to the other PR addressing some PR comments

apps/playground/src/wagmi.ts
```diff
@@ -12,13 +15,24 @@ const {chains, provider, webSocketProvider} = configureChains(
   ],
 );
 
-const {connectors} = getDefaultConnectors({chains});
+const argentConnector: Connector = {
```
Borahm, line 18: I will remove this, but I wanted to see if it works!

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
```diff
@@ -12,20 +13,38 @@ import {ProviderProps} from './types';
...
   orderAttributionMode = 'required',
   statementGenerator,
   theme,
 }: PropsWithChildren<ProviderProps>) => {
   const contextValue: ConnectWalletProviderValue = useMemo(() => {
+    let contextualConnectors = connectors;
+
+    if (!contextualConnectors.length) {
+      const {connectors: defaultConnectors} = buildConnectors({
+        appName: '',
```
jamiely, line 28: Is appName still required?

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
```diff
@@ -12,20 +13,38 @@ import {ProviderProps} from './types';
...
   requireSignature = true,
   orderAttributionMode = 'required',
   statementGenerator,
   theme,
 }: PropsWithChildren<ProviderProps>) => {
   const contextValue: ConnectWalletProviderValue = useMemo(() => {
+    let contextualConnectors = connectors;
+
+    if (!contextualConnectors.length) {
+      const {connectors: defaultConnectors} = buildConnectors({
```
jamiely, line 27: Thoughts on adding a unit test that when `connectors` is empty, `buildConnectors` is called?

jamiely:
I tophatted and, after some code uncommenting, see this:

image

packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
Borahm: Good call. Will add

QuintonC: You might need test-infra for this. IMO I'd defer this addition @bo and add this as a part of the larger feature PR prior to merging into main. Perhaps create a task inside the larger PR to add this unit test.

QuintonC:
> I tophatted and, after some code uncommenting, see this:
> 
> image

To clarify, are you asking if this is an unexpected behavior or stating this is expected? Sorry, just trying to understand this comment :) 

This looks like expected behavior to me since Argent is defined as a WalletConnect connector.

apps/playground/src/assets/Argent.tsx
```diff
@@ -0,0 +1,15 @@
+export const Argent = (
```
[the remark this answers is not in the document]
caropinzonsilva: I'm just commenting here so that we do not forget to remove it :)

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
```diff
@@ -25,18 +24,18 @@ const Template = ({wallets}: {wallets: 'Ethereum' | 'Solana'}) => {
     publicProvider(),
   ]);
 
-  const {connectors} = getDefaultConnectors({chains});
+  const {connectors, wagmiConnectors} = buildConnectors({chains});
```
caropinzonsilva, line 27: This is not a comment for this PR, but just an idea for a future one! We could add a story with custom connectors so that we have somewhere to always test this ✨

caropinzonsilva (approved):
Code looks good! I 🎩 in Chromatic the previous behaviour and it works as expected: https://639b1f308693132693d9b82c-rodvoyslwc.chromatic.com/?path=/story/connect-wallet--playground&args=wallets:Ethereum

QuintonC (commented):
Just a couple of documentation things in the README to address, but otherwise looks solid. Great work 👏

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
Borahm: So, I actually have a question about that. Im not super familiar with Storybooks (on my todo list) and was wondering how you made this work: https://639b1f308693132693d9b82c-rodvoyslwc.chromatic.com/?path=/story/connect-wallet--playground&args=wallets:Ethereum I could not make it work on my local machine and I always got an error.

Aside of that, Im happy to add one, just not 100% sure how to go about it

packages/connect-wallet/src/types/connector.ts
```diff
@@ -65,6 +65,10 @@ export type Connector = Omit<ConnectorInstance, 'createConnector'> & {
   ready?: boolean;
 };
 
+export type CustomConnector = Omit<ConnectorInstance, 'createConnector'> & {
```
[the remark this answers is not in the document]
Borahm: nice!

packages/connect-wallet/src/components/Playground/Playground.stories.tsx
caropinzonsilva: There was an error in that story, but it should be fixed after the Storybook v7 change :)

PR #39, opened 2023-03-01
[analytics] Add `ClientAnalytics` to shared package

caropinzonsilva:
Creates a new shared util `ClientAnalytics` based on Hydrogen's [one](https://github.com/Shopify/hydrogen-v1/blob/main/packages/hydrogen/src/foundation/Analytics/ClientAnalytics.tsx#L15).

This new `ClientAnalytics` util function will allow the developers using the components to subscribe to the tracked events and forward the events to their preferred analytics service. 

You can see in this follow-up example how we can use it, having the Playground as an example: https://github.com/Shopify/blockchain-components/pull/40

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [ ] ~Includes unit tests~ N/A, I added tests, but I noticed the tests under the shared folder are not running. We'll have to address this in a separate PR
- [ ] ~Updated relevant documentation for the changes (if necessary)~ N/A, will update in a follow up PR

### DIFF
--- a/packages/shared/src/utils/analytics/ClientAnalytics.ts
+++ b/packages/shared/src/utils/analytics/ClientAnalytics.ts
@@ -1,0 +1,46 @@
+import {SubscriberFunction, Subscriber, Subscribers} from './types';
+import {eventNames} from './const';
+
+const subscribers: Subscribers = {};
+
+const subscribe = (
+  eventname: string,
+  callbackFunction: SubscriberFunction,
+): Subscriber => {
+  const subs = subscribers[eventname];
+
+  // subs could be undefined, but the casting of subscribers to Subscribers
+  // does not let the linter understand that it could be undefined
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!subs) {
+    subscribers[eventname] = {};
+  }
+
+  const subscriberId = Date.now().toString();
+  subscribers[eventname][subscriberId] = callbackFunction;
+
+  return {
+    unsubscribe: () => {
+      delete subscribers[eventname][subscriberId];
+    },
+  };
+};
+
+const publishEvent = (eventname: string, payload?: any) => {
+  const subs = subscribers[eventname];
+
+  // subs could be undefined, but the casting of subscribers to Subscribers
+  // does not let the linter understand that it could be undefined
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (subs) {
+    Object.keys(subs).forEach((key) => {
+      subs[key](payload);
+    });
+  }
+};
+
+export const ClientAnalytics = {
+  subscribe,
+  publishEvent,
+  eventNames,
+};
--- a/packages/shared/src/utils/analytics/const.ts
+++ b/packages/shared/src/utils/analytics/const.ts
@@ -1,0 +1,3 @@
+export const eventNames = {
+  TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+};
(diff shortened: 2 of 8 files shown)


packages/tokengate/src/index.ts
```diff
@@ -21,3 +21,8 @@ export {Dawn, Default};
  */
 export type {Theme};
 export * from './types';
+
+/**
+ * ClientAnalytics
+ */
+export {ClientAnalytics} from 'shared';
```
caropinzonsilva, line 28: Instead of creating a new package just for the `ClientAnalytics`, I decided to export it from our existing packages. I'll do the same in a follow up PR that adds it to the `connect-wallet`.

packages/shared/src/utils/analytics/ClientAnalytics.ts
```diff
@@ -0,0 +1,46 @@
...
+const subscribe = (
+  eventname: string,
+  callbackFunction: SubscriberFunction,
+): Subscriber => {
+  const subs = subscribers[eventname];
+
+  // subs could be undefined, but the casting of subscribers to Subscribers
+  // does not let the linter understand that it could be undefined
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!subs) {
```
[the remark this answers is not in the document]
jamiely: should the type of subscribers be the below?

```typescript
export type Subscribers = Record<string, Record<string, SubscriberFunction> | undefined>;
```

packages/tokengate/src/components/Tokengate/Tokengate.tsx
```diff
@@ -23,6 +24,13 @@ export const Tokengate = (props: TokengateProps) => {
   } = props;
   const {title, subtitle, sections} = useTokengateCardState(props);
 
+  useEffect(() => {
+    ClientAnalytics.publishEvent(
+      ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
+      {},
```
jamiely, line 30: Can the payload argument be omitted?

jamiely (approved):

packages/shared/src/utils/analytics/ClientAnalytics.ts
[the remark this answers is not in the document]
caropinzonsilva: This removes that error, but then adds 2 new ones 😅 
image

packages/tokengate/src/components/Tokengate/Tokengate.tsx
caropinzonsilva: Good point! I made the payload optional and omitted it from here for now.

packages/tokengate/src/components/Tokengate/Tokengate.tsx
```diff
@@ -23,6 +24,12 @@ export const Tokengate = (props: TokengateProps) => {
   } = props;
   const {title, subtitle, sections} = useTokengateCardState(props);
 
+  useEffect(() => {
+    ClientAnalytics.publishEvent(
+      ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
+    );
```
QuintonC, line 30: Finally got a chance to look at these! So sorry for the delay.

Do you think there's a possibility that we can simplify this by changing the arguments that get passed to `publishEvent` to be either a single event name or a collection of event names? For example:

```ts
    ClientAnalytics.publishEvent(tokengateRendered);
    // OR
    ClientAnalytics.publishEvent([tokengateRendered, connectButtonRendered]);
```

packages/shared/src/utils/analytics/const.ts
```diff
@@ -0,0 +1,3 @@
+export const eventNames = {
+  TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+};
```
QuintonC, line 3: This looks like it is defined more in line with what I would expect to be an enum. Did you intend for this to be an enum? 🤔

PR #40, opened 2023-03-01
[playground] Add AnalyticsListener

caropinzonsilva:
Follow up to https://github.com/Shopify/blockchain-components/pull/39

Shows how the `ClientAnalytics` could be used in an app by adding it to the Playground app.

## 🕹️ Demonstration

01-33-j9egi-9ie2q

- run the playground app
- open console
- see `TOKENGATE_COMPONENT_RENDERED` event in the console only one time per window refresh.

## ✅ Checklist

- [x] Tested on mobile


apps/playground/src/components/AnalyticsListener.tsx
```diff
@@ -0,0 +1,19 @@
+import {useEffect} from 'react';
+import {ClientAnalytics} from '@shopify/tokengate';
+
+export const AnalyticsListener = () => {
+  useEffect(() => {
+    ClientAnalytics.subscribe(
+      ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
+      (payload: any) => {
+        // eslint-disable-next-line no-console
+        console.log(
```
caropinzonsilva, line 10: Here, instead of a console log, the developers can submit the requests to their preferred analytics service.

apps/playground/src/components/AnalyticsListener.tsx
```diff
@@ -0,0 +1,19 @@
+import {useEffect} from 'react';
+import {ClientAnalytics} from '@shopify/tokengate';
+
+export const AnalyticsListener = () => {
+  useEffect(() => {
+    ClientAnalytics.subscribe(
```
jamiely, line 6: It might be good to return `unsubscribe` to improve the example.

jamiely (approved):
:+1: thanks for adding!

apps/playground/src/components/AnalyticsListener.tsx
caropinzonsilva: Done!

apps/playground/src/components/AnalyticsListener.tsx
QuintonC: How does the 3P developer add their service here? Should the ClientAnalytics service instead be running an optional callback?

caropinzonsilva: Instead of the console log, they will track the event however they want. So, for example in the case of [Trekkie](https://github.com/Shopify/trekkie), we could do `window.trekkie.track(params)`

PR #43, opened 2023-03-02
[analytics] Add `ClientAnalytics` tests

caropinzonsilva:
Adds tests for the `ClientAnalytics` util function added in https://github.com/Shopify/blockchain-components/pull/39

## 🕹️ Demonstration
02-31-37572-a7zr5

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


packages/shared/src/utils/analytics/ClientAnalytics.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      mock1,
+    );
+
+    // Force +1 millisecond date update
+    date.setMilliseconds(date.getMilliseconds() + 1);
+    vi.setSystemTime(date);
+    ClientAnalytics.subscribe(
+      ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
+      mock2,
+    );
```
caropinzonsilva, line 46: Whenever we receive a new `subscribe` for the same event, we add it to the array based on the timestamp the subscription was called at: https://github.com/Shopify/blockchain-components/pull/39/files#diff-c5fbe36b19af3837d6fb7cdf9d17c5fa83c8c0f2cec0d8c7b639818065f29fa4R19

This means that there is a small chance that this might not work as expected if two `subscribe` with the same event name are called one after the other one. But I think this is just an edge case, we can worry about it if we receive feedback from other developers that they need this functionality.

jamiely (approved):
Good tests! Two suggestions for possible tests:

* subscriber gets called with specific payload
* subscriber doesn't get called for an event they are not subscribed to (maybe overkill)

packages/shared/src/utils/analytics/ClientAnalytics.test.ts
caropinzonsilva: Makes sense! I'll create an issue for this :)

caropinzonsilva:
@jamiely good suggestions for the tests! I added them in this commit: https://github.com/Shopify/blockchain-components/pull/43/commits/5fa1e8a373788ccd5c098d01aea8eafed1bb3ebb

PR #44, opened 2023-03-02
:bug: Addresses issue with styled-components and webpack

QuintonC:

Addresses an issue with styled-components and webpack. The problem occurs when styled-components is used inside of an ESM only package that is used within a webpack-based bundler (e.g. CRA + Next.js). 

ESM allows for importing in the syntax we're all familiar with. However, the ESM support that Next.js ships with still utilizes `require` instead of `import`. This means that the compiled styled-components code will look like `styled.default.span` rather than `styled.span`, which is why you see the error demonstrated in the before section below.

There are other approaches to resolving this which I considered, but decided against since they require more lift, such as:
- Upgrading to an alpha package of Styled Components (Version 6 is supposed to ship with ESM support, however when testing V6 I ran into type mismatching and other weird behavior.
- Using a different styling library (TailwindCSS, Vanilla Extract, CSS Modules)

To explain the solution here, what I've effectively done is re-exported styled-components with a "hacky" way of providing backwards-compatible CJS/ESM support. [You can read more about this approach here](https://github.com/styled-components/styled-components/issues/3437#issuecomment-1103085056). You can see [this implementation](https://github.com/Shopify/blockchain-components/blob/93a4685cd39eb35143dca5cb5e8e6423205aff8f/packages/shared/src/styles/styled.ts) here.

## 🕹️ Demonstration

| State | Image |
| - | - |
| Before | SCR-20230302-m1h |
| After | SCR-20230302-mz6 |

To test this, you will need a Next.js app to test with

1. Pull this PR
2. Install dependencies, build the packages, and then link the package `dev up && yarn build && cd packages/connect-wallet && yarn link`
3. Navigate to your Next.js test application and run the following: `yarn link "@shopify/connect-wallet"`
4. Run your application and you should see a Wagmi error (temporary)

## ✅ Checklist

- [x] Tested on mobile

### DIFF
--- a/.changeset/loud-cups-promise.md
+++ b/.changeset/loud-cups-promise.md
@@ -1,0 +1,6 @@
+---
+'@shopify/connect-wallet': patch
+'@shopify/tokengate': patch
+---
+
+Addresses an issue with styled components and CJS support. Since the packages support ESM, some frameworks (Next.js) will try to transpile the packages and use a CommonJS version of styled-components we needed to re-export styled-components with a .default key for CJS support.
--- a/packages/connect-wallet/package.json
+++ b/packages/connect-wallet/package.json
@@ -47,7 +47,6 @@
     "redux-logger": "^3.0.6",
     "redux-persist": "^6.0.0",
     "siwe": "^1.1.6",
-    "styled-components": "^5.3.6",
     "ua-parser-js": "^1.0.32"
   },
   "peerDependencies": {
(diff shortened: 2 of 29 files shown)


jamiely (approved):
We've discussed this change offline. I'm good with this change. Tophatted on MetaMask desktop

caropinzonsilva (approved):

PR #48, opened 2023-03-03
Supports allow list using condition desc prop and custom titles

jamiely:

Adds support for customizing what's displayed in segment descriptions in order to support use-cases such as gating using wallet address allow lists.

There are two main changes:

1. `Condition#collectionAddress` is allowed to be optional
2. `Condition#description` has been added and is a React node.
3. All the props of `CustomTitles` may be React nodes

Note that I experimented with `Condition#name` being a React node, but it's expected to be a string in a few different places such as a fragment key. Since `name` doesn't have to be a node to meet the requirements, I left as-is.

### Locked

image

### Unlocked

image

* View (https://gracefulvastphp.jamiely.repl.co/)
* Code (https://replit.com/join/eebzcxrjbv-jamiely)

## ✅ Checklist

- [x] Tested on mobile
- [x] Includes unit tests
- [ ] Updated relevant documentation for the changes (if necessary) - I'll update public docs if this is merged.

### DIFF
--- a/packages/tokengate/src/types.ts
+++ b/packages/tokengate/src/types.ts
@@ -4,7 +4,8 @@ import {ThemeProps} from 'shared';
 export interface Condition {
   name?: string;
   imageUrl?: string;
-  collectionAddress: string;
+  collectionAddress?: string;
+  description?: ReactNode;
 }
 
 export interface Requirements {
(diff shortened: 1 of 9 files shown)


apps/playground/src/preview.ts
```diff
@@ -77,18 +77,11 @@ function getGateRequirement() {
     // Mock API call latency
```
jamiely, line 77: all changes to the playground will be undone before merging this PR

packages/tokengate/src/types.ts
```diff
@@ -2,9 +2,10 @@ import {ReactNode} from 'react';
 import {ThemeProps} from 'shared';
 
 export interface Condition {
```
jamiely, line 4: These types have been widened from `string` to `ReactNode`. Note that ReactNode is a union type that includes `string`.

PR #50, opened 2023-03-06
[analytics] Add onClick tracking to shared Button component

caropinzonsilva:
We would like third-party developers to listen to button's click events so that they can track the user's behaviour with whatever service wanted. 

Instead of adding click events separately to all the buttons, we decided to implement this logic in the shared Button component and control it with the `onClickEventName` prop. If that prop is defined, it means that we want to track the click event of that button. 

## 🕹️ Demonstration

The behaviour of the components is not affected by this change. 

Can be 🎩 using Storybook: https://639b1f308693132693d9b82c-cwcnsconyn.chromatic.com/?path=/story/connect-wallet--playground&args=wallets:Ethereum

## ✅ Checklist

- [x] Tested on mobile
- [x] Includes unit tests


packages/shared/package.json
```diff
@@ -12,6 +12,7 @@
   "types": "src/index.ts",
   "devDependencies": {
     "@storybook/react": "^6.5.13",
+    "@testing-library/react": "^13.4.0",
```
caropinzonsilva, line 15: I had to add this library to the shared packages since I'm adding tests in this PR.

packages/shared/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+            mock={subscriberMock}
+          />
+          <Button {...DefaultArgs} label="label" onClick={onClickMock} />,
+        </>,
+      );
+      const button = await element.findByText('label');
+      fireEvent.click(button);
+      expect(onClickMock).toBeCalledTimes(1);
+      expect(subscriberMock).toBeCalledTimes(0);
+    });
```
caropinzonsilva, line 54: The difference between these 2 tests is that in the second one, I do NOT pass the `onClickEventName` to the `Button`, so the `subscriberMock` is not called.

packages/shared/src/utils/test.ts
```diff
@@ -0,0 +1,9 @@
+import {render, RenderOptions} from '@testing-library/react';
+import {ReactElement} from 'react';
+
+import {RootProvider} from '../providers/RootProvider';
+
+export const renderWithContext = (
+  ui: ReactElement,
+  options?: Omit<RenderOptions, 'wrapper'>,
+) => render(ui, {wrapper: RootProvider, ...options});
```
caropinzonsilva, line 9: In the future, we can consider also adding the `I18nextProvider` so that we could remove the `tokengate` specific test utils: https://github.com/Shopify/blockchain-components/blob/main/packages/tokengate/src/tests/test-utils.tsx#L17

But I think that change is unrelated to this PR.

packages/shared/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    it('sends onClick event if onClickEventName is defined', async () => {
+      const onClickMock = vi.fn();
+      const subscriberMock = vi.fn();
+      const onClickEventName = 'onClickEventName';
+      const element = renderWithContext(
+        <>
+          <AnalyticsListenerTestHelper
+            eventName={onClickEventName}
+            mock={subscriberMock}
+          />
```
caropinzonsilva, line 21: This is a new util component that I added in this PR that will make testing the analytics listeners easier: https://github.com/Shopify/blockchain-components/pull/50/files#diff-c1e5a8e07f3df594572f11d9027050a5f671e25b54407d2d2a30ca3b24faf5c0

packages/shared/src/components/Button/types.ts
```diff
@@ -9,6 +9,7 @@ interface ButtonBaseProps {
   primary?: boolean;
   disabled?: boolean;
   size?: Size;
+  onClickEventName?: string;
```
jamiely, line 12: did you consider either overloading this to be `onClickEventNameOrPayload` or having an alternate payload attribute? Probably out of scope for this issue.

packages/shared/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import {Button} from './Button';
+import {DefaultArgs} from './fixtures';
+
+describe('Button', () => {
+  describe('analytics', () => {
+    it('sends onClick event if onClickEventName is defined', async () => {
+      const onClickMock = vi.fn();
+      const subscriberMock = vi.fn();
+      const onClickEventName = 'onClickEventName';
```
jamiely, line 15: If the test is a form of documentation, would it be better for this event name to be something like `CONNECT_BUTTON_CLICKED` to further illustrate the purpose?

packages/shared/src/utils/analytics/helpers/test.tsx
```diff
@@ -0,0 +1,20 @@
...
+  mock,
+}: {
+  eventName: string;
+  mock: (payload: any) => void;
+}) => {
+  useEffect(() => {
+    const {unsubscribe} = ClientAnalytics.subscribe(eventName, (payload: any) =>
+      mock(payload),
+    );
+    return () => unsubscribe();
```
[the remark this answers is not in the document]
caropinzonsilva: Done! Thanks for the suggestion

packages/shared/src/components/Button/Button.test.tsx
caropinzonsilva: You are right! I changed it, thank you

packages/shared/src/components/Button/types.ts
caropinzonsilva: Good question about the payload 🤔 I didn't really consider it. So far, the click events that we have do not require any payload. Is it ok if we leave it as is and then consider adding that new prop if we ever need to add the payload? That way, we avoid adding things that we do not need yet.

jamiely: sure!

packages/shared/src/components/Button/Button.stories.tsx
```diff
@@ -4,7 +4,7 @@ import {ComponentStory, ComponentMeta, Parameters} from '@storybook/react';
 import {Text} from '../Text';
 
 import {Button} from './Button';
-import type {ButtonProps} from './types';
+import {DefaultArgs} from './fixtures';
```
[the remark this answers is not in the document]
QuintonC: There might be a merge conflict here with the new Storybook PR (just a heads up).

packages/shared/src/utils/test.ts
QuintonC: I do think it's nice to keep test utils + test context functionality inside of a `test` directory with in the source code so that test code is not accidentally imported while importing files. I think this is a common practice in most of our other work as well.

Could we move the test helpers + test context here to a test dir [similar to what is seen in the `connect-wallet` package](https://github.com/Shopify/blockchain-components/tree/main/packages/connect-wallet/src/test)?

image

packages/shared/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+          ,
+        </>,
+      );
+      const button = await element.findByText('label');
+      fireEvent.click(button);
+      expect(onClickMock).toBeCalledTimes(1);
+      expect(subscriberMock).toBeCalledTimes(1);
+    });
+
+    it('does not sends onClick event if onClickEventName is undefined', async () => {
```
QuintonC, line 38:
```suggestion
    it('does not send onClick event if onClickEventName is undefined', async () => {
```

packages/shared/src/components/Button/Button.tsx
```diff
@@ -12,8 +15,18 @@ export const Button = ({
...
+  onClickEventName,
   ...props
 }: ButtonProps) => {
+  const onClickWithTracking = useCallback(() => {
+    if (onClickEventName) {
+      ClientAnalytics.publishEvent(onClickEventName);
+    }
+    if (onClick) {
+      onClick();
+    }
```
QuintonC, line 28: Could you do `props?.onClick()` here?

caropinzonsilva: `props` is always undefined, so the first `?` is unnecessary. It would have to be something like `props.onClick?.()`
The other reason why I did it that way is so that we only include the `onClick` variable in the dependency array of the `useCallback`. If we want to use the `props`, we would need to have all the props as a dependency. So also a minor performance improvement.

packages/shared/src/components/Button/Button.stories.tsx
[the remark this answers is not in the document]
caropinzonsilva: Resolved!

packages/shared/src/utils/test.ts
caropinzonsilva: Done! https://github.com/Shopify/blockchain-components/pull/50/commits/aa03548881812b18b7a862e9886c5a9989e3175e

PR #51, opened 2023-03-06
Sample concept for simplified type names

QuintonC:
⚠️ Fixes: 

...

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/shared/src/utils/analytics/types.ts
```diff
@@ -1,7 +1,13 @@
...
+export type EventName = `${Event}`;
+
 export interface Subscriber {
   unsubscribe: () => void;
 }
 
 export type SubscriberFunction = (payload: any) => void;
 export type SubscriberValue = Map<string, SubscriberFunction>;
-export type Subscribers = Map<string, SubscriberValue>;
+export type Subscribers = Map<EventName | string, SubscriberValue>;
```
jamiely, line 13: I would expect that the key can only be an `EventName`. Why is it allowed to be a string here?

jamiely (approved):
I'm good with this from a technical perspective. 

One consideration with this change is it may make it harder for people to configure analytics, since they will either have to cast `any` or use a different analytics method (which they will probably do anyway).

Before this (I'm assuming) someone could theoretically use our analytics package throughout their app.

QuintonC:
> I'm good with this from a technical perspective. 
> 
> 
> 
> One consideration with this change is it may make it harder for people to configure analytics, since they will either have to cast `any` or use a different analytics method (which they will probably do anyway).
> 
> 
> 
> Before this (I'm assuming) someone could theoretically use our analytics package throughout their app.

I mean in theory they could use their analytics package in the callback to re-dispatch the event to their own service 🤔

But I think my assumption for the analytics work was that it would be for component based events not gate-based events or anything like that. For example, the connect-wallet package would never have context of the gates. So it doesn't make sense that it would do any sort of gate-based analytics. 

My thoughts are that if we want things that granular they can be added in callbacks to an analytics service or they can be requested via feature request. 

What do you think?

jamiely:
> I mean in theory they could use their analytics package in the callback to re-dispatch the event to their own service 🤔

Yes, this may be harder than just using ours.

> What do you think?

I think this is more of a product decision and I don't have a particular viewpoint on it.

PR #56, opened 2023-03-07
[analytics] Refactor AnalyticsClient to be a Provider

caropinzonsilva:
With the architecture that we had before (importing a file with util functions), we were having issues with having multiple instances of the file. This caused the subscribers not getting triggered 

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests

### DIFF
--- a/apps/playground/src/components/AnalyticsListener.tsx
+++ b/apps/playground/src/components/AnalyticsListener.tsx
@@ -1,5 +1,5 @@
 import {useContext, useEffect} from 'react';
-import {AnalyticsContext} from '@shopify/blockchain-components-analytics';
+import {AnalyticsContext} from 'shared';
 
 export const AnalyticsListener = () => {
   const {subscribeToAll} = useContext(AnalyticsContext);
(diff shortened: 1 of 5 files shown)


packages/shared/src/providers/AnalyticsProvider/utils.test.ts
```diff
@@ -0,0 +1,92 @@
+import {vi} from 'vitest';
```
caropinzonsilva, line 1: Copy + paste from: https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/utils/analytics/ClientAnalytics.test.ts

packages/shared/src/providers/AnalyticsProvider/const.ts
```diff
@@ -0,0 +1,3 @@
+export const eventNames = {
+  TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+};
```
caropinzonsilva, line 3: Copy + paste from: https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/utils/analytics/const.ts

PR #62, opened 2023-03-08
[analytics] create new internal package 'analytics'

caropinzonsilva:
Create a new internal package `analytics` and move the [ClientAnalytics](https://github.com/Shopify/blockchain-components/tree/main/packages/shared/src/utils/analytics) from the `shared` package into this new package. 

## Why
With the previous implementation, the events published in the shared `Button` were not captured by the subscribers.

After some investigation, we noticed that the file [ClientAnalytics](https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/utils/analytics/ClientAnalytics.ts) was getting imported twice. This created two versions of the subscribers which caused the bug.

### Other solutions
We tried these other solutions, but none of them worked:
- Importing the`ClientAnalytics` from the shared package instead of re-exporting them from `@shopify/tokengate`
- Converting the utils into a provider and exporting them from the `shared` package: https://github.com/Shopify/blockchain-components/pull/56
- Converting the utils into a redux provider and exporting them from the `shared` package: https://github.com/Shopify/blockchain-components/pull/61

The constant in the failed solutions that we saw was the `shared` package. Our theory is that, since the `Button` is also in the `shared` package, it creates a duplicate of the functions/providers that causes the issue.

## 🕹️ Demonstration
08-00-cpw58-dr8im

NOTE: there is a duplicated tokengate render event that I will address in another PR.

Can be 🎩 in https://replit.com/@caropinzonsilva/WoozyGiddyMonotone

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


apps/playground/package.json
```diff
@@ -27,6 +27,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@shopify/blockchain-components-analytics": "*",
```
caropinzonsilva, line 30: **Note for the reviewer**
In order to facilitate the reviewing, I split this PR into 2 commits:
1. Creation of the new package: https://github.com/Shopify/blockchain-components/pull/62/commits/b41679b23c40e357c8bfab10e08ef69fa5e8a3d7
2. Movement of the `ClientAnalytics` from `shared` package into new one:mhttps://github.com/Shopify/blockchain-components/pull/62/commits/99c4dd57eb7140ece971e50cb7869573c4a4887d

apps/playground/src/components/AnalyticsListener.tsx
```diff
@@ -1,18 +1,12 @@
...
-      ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
-      (payload: any) => {
-        // eslint-disable-next-line no-console
-        console.log(
-          ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
-          payload,
-        );
-      },
-    );
+    const {unsubscribe} = subscribeToAll((payload: any) => {
```
caropinzonsilva, line 6: Instead of just subscribing to one event, we are now subscribing to all in order to make development and testing easier.

packages/blockchain-components-analytics/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@shopify/blockchain-components-analytics",
+  "version": "0.0.1",
+  "private": true,
```
caropinzonsilva, line 4: I made this package `private` for now.

apps/playground/src/components/AnalyticsListener.tsx
```diff
@@ -1,18 +1,12 @@
...
-        console.log(
-          ClientAnalytics.eventNames.TOKENGATE_COMPONENT_RENDERED,
-          payload,
-        );
-      },
-    );
+    const {unsubscribe} = ClientAnalytics.subscribeToAll((payload: any) => {
+      // eslint-disable-next-line no-console
+      console.log('Captured event', payload);
+    });
```
caropinzonsilva, line 9: Instead of just subscribing to one event, we are now subscribing to all in order to make development and testing easier.

packages/shared/src/components/Button/Button.tsx
```diff
@@ -1,8 +1,8 @@
 import {useCallback} from 'react';
+import {ClientAnalytics} from '@shopify/blockchain-components-analytics';
```
QuintonC, line 2: Would this make the individual packages (e.g. Tokengate + Connect Wallet) dependent on `@shopify/blockchain-components-analytics`?

caropinzonsilva: Yeah, they would depend on it as they depend on the `shared` package.

caropinzonsilva: Initially, this new package `@shopify/blockchain-components-analytics` is only private, so we would want it to work the same way as the `shared` package. 

I was hoping that this would happen by just making private the package, but I can look into it further.

QuintonC: > I want it to work exactly the same as the `shared` package, so instead of it being an external dependency, it gets compiled (as I think you mentioned in a previous comment)

Awesome! Thanks for the context, that makes more sense.

What are your thoughts on shortening the package name since this will be similar to shared and not published? Perhaps just `analytics`?

apps/playground/src/index.tsx
```diff
@@ -43,6 +43,7 @@ window.playground = {
               chains={chains}
               orderAttributionMode="ignoreErrors"
             >
+              <AnalyticsListener />
```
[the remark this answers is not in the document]
jamiely: Good as-is. Does the analytics listener need to be inside the `<ConnectWalletProvider/>`? If not, we may want to move it higher in order to indicate that the provider is not required.

packages/shared/src/components/Button/Button.tsx
jamiely: If I didn't have context on the project, I think that the long name would suggest to me that the package is public, so I am in favor of changing the name to analytics.

jamiely (approved):
Looks good!

apps/playground/src/index.tsx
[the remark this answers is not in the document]
caropinzonsilva: It does not! I'll move it up, good point.

packages/shared/src/components/Button/Button.tsx
caropinzonsilva: Package renamed! https://github.com/Shopify/blockchain-components/pull/62/commits/f02f5d309435cc6d0b01ff5f6ea59ad2a0b2ff38

apps/playground/src/components/AnalyticsListener.tsx
```diff
@@ -1,18 +1,12 @@
 import {useEffect} from 'react';
-import {ClientAnalytics} from '@shopify/tokengate';
+import {ClientAnalytics} from 'analytics';
```
QuintonC, line 2: If we're not exporting the package publicly should this be imported from either `@shopify/tokengate` or `@shopify/connect-wallet`?

apps/playground/package.json
```diff
@@ -30,6 +30,7 @@
     "@shopify/gate-context-client": "*",
     "@shopify/connect-wallet": "*",
     "@shopify/tokengate": "*",
+    "analytics": "*",
```
QuintonC, line 33: I think we can remove this since it's not a package that can be imported by 3P developers.

babel.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
     './shared',
     './tokengate',
     './gate-context-client',
+    './blockchain-components-analytics',
```
QuintonC, line 16: I thought we had gotten rid of babel entirely since we're using `Vite`, `Vitest` and `tsup` 🤔 

I wonder if we can remove this file.

packages/connect-wallet/package.json
```diff
@@ -37,6 +37,7 @@
     "@reduxjs/toolkit": "2.0.0-alpha.2",
     "@shopify/gate-context-client": "*",
     "@storybook/react": "^7.0.0-beta.52",
+    "analytics": "*",
```
QuintonC, line 40: If we're utilizing this package the same way as `shared` then we have to [resolve the package via `tsconfig` paths](https://github.com/Shopify/blockchain-components/blob/main/packages/connect-wallet/tsconfig.json#L5) instead of adding it to dependencies since [tsup doesn't bundle modules listed in dependencies](https://tsup.egoist.dev/#excluding-packages).

> By default tsup bundles all `import`-ed modules but `dependencies` and `peerDependencies` in your packages.json are always excluded, ...

packages/tokengate/package.json
```diff
@@ -40,6 +40,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "analytics": "*",
```
QuintonC, line 43: Related comment (will edit this one with a link after review is finished)

QuintonC (commented):
Looking good! Just a few things of note with the context of how we're using the analytics package.

apps/playground/src/components/AnalyticsListener.tsx
caropinzonsilva: I did this initially, but this causes the bug unfortunately, so it needs to be imported from the package.

caropinzonsilva:
Closing this PR. I created a clean version of it here: https://github.com/Shopify/blockchain-components/pull/74

PR #63, opened 2023-03-08
[analytics] export functions instead of objects

caropinzonsilva:
Small refactor to export directly the functions instead of exporting an object. This will make the function calls less verbose.

Green ci 🍏 

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


lesliexin (approved):
Nice change, looks so much cleaner! :)

PR #65, opened 2023-03-08
[analytics] Create new hook `useEventWithTracking`

caropinzonsilva:
Create a new hook `useEventWithTracking`  that can be used in components in which we want to track the click events. I refactored the `Button` component to use this new hook as an example.

## Why
Initially, we only added the button tracking to our shared `Button` component: https://github.com/Shopify/blockchain-components/pull/50

However, while I was adding buttons to track I realized that not all our buttons use that shared component. Here are some buttons that we want to track but do not use internally the `Button` component:
- [ConnectorButton](https://github.com/Shopify/blockchain-components/blob/main/packages/connect-wallet/src/components/ConnectorButton/ConnectorButton.tsx)
- [IconButton](https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/components/IconButton/IconButton.tsx)
- [ConnectedButton](https://github.com/Shopify/blockchain-components/blob/main/packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx#L67)

## 🕹️ Demonstration
08-48-fv0qw-kzxaz

- Run `yarn dev`
- Open playground in the browse + console
- When we click the `connect wallet` button, an event should be captured in the console.

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


packages/blockchain-components-analytics/src/ClientAnalytics/utils.test.tsx
```diff
@@ -89,4 +97,63 @@ describe('utils', () => {
...
+          />
+          <Button onClick={onClickMock} />,
+        </>,
+      );
+      const button = await element.findByText('button text');
+      fireEvent.click(button);
+      expect(onClickMock).toBeCalledTimes(1);
+      expect(subscriberMock).toBeCalledTimes(0);
+    });
+  });
```
caropinzonsilva, line 158: These tests are really similar to the ones we already have in the `Button` component: https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/components/Button/Button.test.tsx

I decided to keep both to have better test coverage (both in the hook and in a component).

lesliexin (approved):
Tophatted and works as expected!

packages/blockchain-components-analytics/src/ClientAnalytics/utils.ts
```diff
@@ -77,3 +79,30 @@ export const publishEvent = (eventname: string, payload?: any) => {
...
   subscriptions?.forEach((subscription) => subscription(payload));
 };
+
+/**
+ * @param {() => void | undefined)} onClick The original onClick function
+ * @param {string} eventname An event name indicating which event to publish to subscribers.
+ * @param {any | undefined} payload The data which was dispatched to the event.
+ * @returns {() => void}
+ */
+export const useOnClickWithTracking = ({
```
jamiely, line 89: Good as-is. Should this be a more generic `useEventWithTracking`?

packages/blockchain-components-analytics/src/ClientAnalytics/utils.test.tsx
```diff
@@ -89,4 +97,63 @@ describe('utils', () => {
       expect(mock1).toHaveBeenCalledTimes(0);
     });
   });
+
+  describe('useOnClickWithTracking', () => {
```
jamiely, line 101: Good as-is. I'd consider testing whether `useOnClickWithTracking` executes `publishEvent` as well.

jamiely (approved):

packages/blockchain-components-analytics/src/ClientAnalytics/utils.test.tsx
caropinzonsilva: Although it is not explicit, this is already covered in our tests. We are mounting the `AnalyticsListenerTestHelper` that subscribes to the event published: https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/tests/helpers/ClientAnalytics.tsx#L12

Since we are testing that the `subscriberMock` gets called, we are implicitly testing that the `publishEvent` also happened. More like an integration test, rather than testing directly that the `publishEvent` is executed.

packages/blockchain-components-analytics/src/ClientAnalytics/utils.ts
caropinzonsilva: Done! https://github.com/Shopify/blockchain-components/pull/65/commits/9dd59b4d86209f5572c2566792638e5dbcab1840

Good suggestion, in case we need to track another event in the future

PR #67, opened 2023-03-08
Adding test infra for providers and testing useConnectorData

lesliexin:

To be able to test the changes made to the `ConnectWalletProvider` in the [Customizable wallet list](https://github.com/Shopify/blockchain-components/pull/55) epic, we need to:
1. Add test infrastructure to be able to test with contexts and test hooks with context. This is also outlined in [this issue](https://github.com/Shopify/nft-gating/issues/1315) and this PR will close it as well. 
2. Test a consumer of `ConnectWalletProvider`, which in this case is the `useConnectorData` hook

## ✅ Checklist

- [x] ~~Tested on mobile~~
- [x] Includes unit tests

### DIFF
--- a/packages/connect-wallet/src/hooks/useConnectorData.test.ts
+++ b/packages/connect-wallet/src/hooks/useConnectorData.test.ts
@@ -1,0 +1,41 @@
+import {
+  CHAINS,
+  CUSTOM_CONNECTOR,
+  DEFAULT_CONNECTOR_COLLECTION,
+} from '../test/fixtures/connector';
+import {createWagmiFixture} from '../test/fixtures/wagmi';
+
+import {renderHookWithContext} from '../test/utils';
+
+import {useConnectorData} from './useConnectorData';
+
+describe('useConnectorData', () => {
+  it('returns custom connector id', () => {
+    const {client} = createWagmiFixture({customConnectors: [CUSTOM_CONNECTOR]});
+
+    const {result} = renderHookWithContext(
+      () => useConnectorData({id: CUSTOM_CONNECTOR.id}),
+      {
+        chains: CHAINS,
+        client,
+        connectors: [CUSTOM_CONNECTOR],
+      },
+    );
+
+    expect(result.current).toEqual(expect.objectContaining(CUSTOM_CONNECTOR));
+  });
+
+  it('returns correct data based on id for default connectors', () => {
+    const {client, connectors} = createWagmiFixture({});
+
+    DEFAULT_CONNECTOR_COLLECTION.forEach((id) => {
+      const {result} = renderHookWithContext(() => useConnectorData({id}), {
+        chains: CHAINS,
+        client,
+        connectors,
+      });
+
+      expect(result.current).toEqual(expect.objectContaining({id}));
+    });
+  });
+});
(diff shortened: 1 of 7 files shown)


packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
```diff
@@ -11,15 +11,15 @@ import store from '../../store/configureStore';
 import {ConnectWalletContext, ConnectWalletProviderValue} from './context';
 import {ProviderProps} from './types';
 
-export const ConnectWalletProvider: FC<PropsWithChildren<ProviderProps>> = ({
```
lesliexin, line 14: Removing `PropsWithChildren` as `ProviderProps` already supports it

packages/connect-wallet/src/test/fixtures/wagmi.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type WagmiFixtureProps = Omit<BuildConnectorsProps, 'chains'> & {
+  chains?: Chain[];
+};
+
+export const createWagmiFixture = ({
+  appName,
+  chains = [mainnet],
+  customConnectors,
+  excludedConnectors = ['coinbaseWallet'],
```
lesliexin, line 15: Excluding Coinbase by default because it throws the following error: ` this.WebSocketClass is not a constructor` which hints that we need to mock it. Since the logic for the hook we're testing is the same for each connector, I'm comfortable excluding it here.

packages/connect-wallet/src/test/fixtures/connector.ts
```diff
@@ -6,7 +6,7 @@ import {publicProvider} from 'wagmi/providers/public';
 import {SerializedConnector} from '../../types/connector';
 
 export const DEFAULT_SERIALIZED_CONNECTOR: SerializedConnector = {
-  id: 'MetaMask',
+  id: 'metaMask',
```
lesliexin, line 9: Realized that the id was wrong

packages/connect-wallet/src/test/fixtures/wagmi.ts
```diff
@@ -0,0 +1,36 @@
+import {buildConnectors} from 'src';
+import {BuildConnectorsProps} from 'src/connectors/buildConnectors';
```
QuintonC, line 2: Can we update the paths for these?

QuintonC (approved):
It was great pairing with you and watching you piece this together. Excellent work 👏

packages/connect-wallet/src/hooks/useConnectorData.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    DEFAULT_CONNECTOR_COLLECTION.forEach((id) => {
+      const {result} = renderHookWithContext(() => useConnectorData({id}), {
+        chains: CHAINS,
+        client,
+        connectors,
+      });
+
+      expect(result.current).toEqual(expect.objectContaining({id}));
+    });
```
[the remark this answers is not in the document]
lesliexin: Ooh what do you mean by connector ordering? (Sorry I might be missing something)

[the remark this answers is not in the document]
lesliexin: `useConnectorData` only returns one connector at a time based on a single `id` parameter so we wouldn't be able to test ordering.

PR #74, opened 2023-03-09
@shopify/blockchain components package

caropinzonsilva:
Create a new internal package `@shopify/blockchain-components` and move the [ClientAnalytics](https://github.com/Shopify/blockchain-components/tree/main/packages/shared/src/utils/analytics) from the `shared` package into this new package. 

## Why
With the previous implementation, the events published in the shared `Button` were not captured by the subscribers.

After some investigation, we noticed that the file [ClientAnalytics](https://github.com/Shopify/blockchain-components/blob/main/packages/shared/src/utils/analytics/ClientAnalytics.ts) was getting imported twice. This created two versions of the subscribers which caused the bug.

### Other solutions
We tried these other solutions, but none of them worked:
- Importing the`ClientAnalytics` from the shared package instead of re-exporting them from `@shopify/tokengate`
- Converting the utils into a provider and exporting them from the `shared` package: https://github.com/Shopify/blockchain-components/pull/56
- Converting the utils into a redux provider and exporting them from the `shared` package: https://github.com/Shopify/blockchain-components/pull/61

The constant in the failed solutions that we saw was the `shared` package. Our theory is that, since the `Button` is also in the `shared` package, it creates a duplicate of the functions/providers that causes the issue.

## 🕹️ Demonstration
08-00-cpw58-dr8im

NOTE: there is a duplicated tokengate render event that I will address in another PR.

Can be 🎩 in https://replit.com/@caropinzonsilva/WoozyGiddyMonotone

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


.changeset/giant-bees-scream.md
```diff
@@ -0,0 +1,5 @@
+---
```
caropinzonsilva, line 1: **Note for the reviewer**
In order to make the reviewing process easier, I split the changes in 2 commits:
1. Scaffold of adding a new package: [[analytics] create new package @shopify/blockchain-components](https://github.com/Shopify/blockchain-components/pull/74/commits/d3b80f40d2f7f667d02d08507abe25f8234a18f1)
2. Moving the ClientAnalytics to the new package: https://github.com/Shopify/blockchain-components/pull/74/commits/2ead93b4c0ab2870f13776a1e0971451b4aeda97

packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
```diff
@@ -1,3 +1,4 @@
 export const eventNames = {
   TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+  CONNECT_BUTTON_CLICKED: 'CONNECT_BUTTON_CLICKED',
```
caropinzonsilva, line 3: I'm adding this new event name so that we can test the event clicks.

.changeset/giant-bees-scream.md
```diff
@@ -0,0 +1,5 @@
...
+'@shopify/blockchain-components': patch
+---
+
+Initial release
```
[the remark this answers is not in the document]
QuintonC: Is this still relevant? I'm seeing two different changesets (I think there should only be one). You might want to rebase this on main if you haven't done so recently.

babel.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
     './shared',
     './tokengate',
     './gate-context-client',
+    './blockchain-components',
```
QuintonC, line 16: Just so we don't lose track of this -- can we create an issue for the babel file since we _might_ not need it anymore?

[Related](https://github.com/Shopify/blockchain-components/pull/62#discussion_r1131377062)

.changeset/giant-bees-scream.md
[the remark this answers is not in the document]
caropinzonsilva: Both of the patches are still relevant, I can combine them into one if desired but it does would not make a difference at the end.

babel.config.js
caropinzonsilva: Sure!

caropinzonsilva: Added to the Ideas column: https://github.com/Shopify/nft-gating/issues/1322

PR #75, opened 2023-03-09
Showing custom or marketplace links in a modal on the tokengate component

Borahm:
⚠️ Fixes: https://github.com/Shopify/nft-gating/issues/1268

## What

Screenshot 2023-03-22 at 18 07 49

This adds support for linking to an external source (e.g. Marketplaces) for a buyer to purchase an unlocking token if they don't have one.

Developers can pass an array of custom URLs with link text and supporting favicons to the component. If developers don't pass a favicon image, we have a link icon fallback for the most common marketplaces or a generic icon

## ✅ Checklist

- [x] ~~Tested on mobile~~
- [x] Includes unit tests
- [x] ~~Updated relevant documentation for the changes (if necessary)~~ POST-MERGE


packages/tokengate/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <a
+      href={url}
+      target="_blank"
+      rel="noreferrer"
+      className="sbc-flex sbc-w-full sbc-flex-row sbc-gap-x-4 sbc-no-underline"
+    >
+      <div className="sbc-h-5 sbc-w-5 sbc-text-center">{iconElement}</div>
+      <Text color="primary">
+        {marketplace ? t('view', {link: marketplace}) : label}
```
QuintonC, line 59:
```suggestion
        {marketplace ? t('view', {marketplace}) : label}
```

packages/shared/src/components/Popover/Popover.tsx
```diff
@@ -52,7 +59,7 @@ export const Popover = ({
       <AnimatePresence>
         {visible ? (
           <m.div
-            className="sbc-fixed sbc-top-0 sbc-bottom-0 sbc-right-0 sbc-z-max sbc-flex sbc-min-w-full sbc-flex-col sbc-justify-end sbc-p-0 sm:sbc-absolute sm:sbc-top-full sm:sbc-bottom-auto sm:sbc-min-w-[280px] sm:sbc-py-1"
+            className={`sbc-fixed sbc-top-0 sbc-bottom-0 sbc-right-0 sbc-z-max sbc-flex sbc-min-w-full sbc-flex-col sbc-justify-end sbc-p-0 sm:sbc-absolute sm:sbc-top-full sm:sbc-bottom-auto sm:sbc-py-1 ${containerClass}`}
```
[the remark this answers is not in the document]
Borahm: We replaced `min-w ` with the `containerClass` prop, because the `ConnectButton` and `TokenBase` components require different `min-w` to make it look right in both cases. The `ConnectButton` requires 280px vs 150px for the `TokenBase` component.

packages/shared/src/icons/Blur.tsx
```diff
@@ -0,0 +1,27 @@
...
...MOXX+IBdGhVuZDVKa8QRX17ziEMFC3eMWYirD6uFrnp1zYQd7+t1B49bb0iFujFvY7w+JWArark1YfXI1GqKVP+wL3f1bbyeFQK+GSjAaqqfZcCLgrfRU6Nsr2ERUHHzXYps64/SGZsmohLjtQPLiQUVWepRi5vy+JYZGF1w/iAXTBVuZBq414iHfXvOIQ6oWN4zTN6s1/uxq1X4zz/AGoE1abuQG9ktpPLUQJdIKK8/Ma6Npteb8RK+WtELhwr4mgj5AQKtuX9PEBAKBXHf3lCqADzhGgMAl41lWdTSrqKiQrWPiB228irAlnQndyFEqourz7ThEs9fxAGbZSP7f5lG0XuK5yWXQt3r45+Yq/I2vxFUyMtgAPbzOTT2ss1/uxq1X4zi8/xCmrTdyArsl4nlqINKQAV5iDRtNrzfiJ8/QL/AEhKDhXxNBPYCBV3y/OeIBEoHp395QrAXzh6jUDAJeNZVi4dKuokJDdY+IFb8xVgSzoTu5AmBRdXn2nkVfVePiAM2ykf2/zKOJe4rnItqFu9fHJTeWNqK9LygAPD4yCda5ZvzGC72k/v7fpMlhEulGCa6KMl5RrpqAJRl6Sh04ZcFqpPgVFsDTucJ5KJzTxEoG64zAopjvv9J0p13B2KVLo+/YgA89RQWl8ksQPvccLQ0/qgXITUpoTYUbFu2eLyaYHChC72JSgUq1i0LL7xGlgXnXktaU998lktoyzflhQvqq9/3xMlhKulGCtPSjJbw0dagCZl6Sh2MMuC0knwKiUDTucInaJzTxEpDdcZQKKY77/SYqFfcHYtUuj79IgA4epYL+qJY0fe5wtDv6oFyE1KaEmGx7tni7JpAcKELvYiFAeWsWiyf1uNaWBfy6S1pT33yLRoltslorNqnxHeEuqDUuOLF3SeZon7XiBSqA9Qq7S98V9400YaS1aCOvEsgTd1KYW10MlOHeUUwUUmt3fYonD2rLqzXuyv2gLQHguC6wSzfNQLK5XexhjZ7Yaw/p7ijj5NRxa1KnoghCF2wOnzBsUj0i0aJ1slorN4nNjrxLqg1LhxYu6TzNE/Y8QKVQHqFXaPfFfeJNGGkVWBF3xLNE3dSmja6cyIcO8opgopWbu+xRJQ+VZZWa92V+0BUgPBcTrBM3yECyuV3sYY2PljAhr49xR+qzUSlrUqej1BLgXbEe3VhstTyJ//2Q=="
```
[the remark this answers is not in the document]
Borahm: Yeah, I had the same comment. I just got a new one without the jpeg inside

apps/playground/src/App.tsx
```diff
@@ -6,6 +6,7 @@ import {
 } from '@shopify/tokengate';
 
 import './App.css';
+import type {Link} from '@shopify/tokengate';
```
QuintonC, line 9: This import should be moved up (this violates ESLint)

packages/tokengate/src/components/TokenBase/TokenBase.tsx
```diff
@@ -1,59 +1,116 @@
...
 
-export {TokenBase};
+        {links?.length && !isUnlocked ? (
+          <IconButton
+            aria-label={t('more') as string}
+            icon={More}
+            onClick={togglePopover}
+          />
+        ) : (
+          rightContent
```
jamiely, line 99: We talked about this but it seems weird to me that the links are only visible if we are locked.

packages/tokengate/src/components/TokenBase/TokenBase.tsx
```diff
@@ -1,59 +1,116 @@
...
+              color="secondary"
+              variant="bodyMd"
+            >
+              {subtitle}
+            </Text>
+          ) : null}
+        </div>
 
-export {TokenBase};
+        {links?.length && !isUnlocked ? (
```
jamiely, line 92: This is good as-is. 

A suggestion for a different way of implementing this would be to pass empty or null links to `TokenBase` if we are unlocked. This would avoid the component needing to know about locking/unlocking, and possible avoid other refactoring. 

I'd say the way you have done it is better if we really want this component to know about locking/unlocking.

caropinzonsilva (approved):
The code looks good! I just left a comment related to removing the `isUnlocked` prop from the `TokenBase`, but it is non blocking

packages/tokengate/src/components/TokenBase/TokenBase.tsx
Borahm: @caropinzonsilva brought a good point. `isUnlocked` is actually not necessary, since `isUnlocked` is only passed in the conditions and the conditions are only are only displayed when the component is locked.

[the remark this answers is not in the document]
Borahm: that makes sense! Thannk you

apps/playground/src/App.tsx
QuintonC: > Specifically this should be merged with the previous import from `@shopify/tokengate`

Importing the type separate from the regular imports is normal.

[ESLint source](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md)

PR #78, opened 2023-03-10
[analytics] Add Shopify service attribute to all published events

caropinzonsilva:
We want to send the Shopify Service (pdp, checkout, cart, ...) as an attribute related to a published event. 

I'm creating a new `getShopifyService` util function that analyses the URL's pathname and identifies which Shopify Service it corresponds to. In case the URL does not match the typical structure, we send `OTHER` as the attribute. This might not work for headless stores that define their own URL structure, but it is a known edge case that we accept.

I tried adding good test coverage for this new function. 
- Green CI 🍏 

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.ts
```diff
@@ -106,3 +109,28 @@ export const useEventWithTracking = ({
...
+  return {
+    shopifyService: getShopifyService(pathname),
+  };
+};
+
+/**
+ * @param {string} pathname The client's complete URL's pathname
+ * @returns {string} A string that represents the Shopify Service the URL corresponds to (Checkout, PDP, other)
+ */
+export const getShopifyService = (pathname: string) => {
```
caropinzonsilva, line 129: Initially, I was going to use the `window.Shopify` object that we use in the util [isShopifyStore](https://github.com/Shopify/blockchain-components/blob/main/packages/gate-context-client/src/cartAjaxApi/shopify.ts#L3).

However, I realized that some shops (Gymshark, for example) do not use this variable, so I decided to base the implementation in the URL in order to increase the coverage of the solution.

packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.ts
```diff
@@ -106,3 +109,28 @@ export const useEventWithTracking = ({
     },
     [callback, eventName, payload],
   );
+
+/**
+ * @returns {object} An object containing the additional payload we want to send to all published events
+ */
+export const getAdditionalEventPayload = () => {
+  // Casting as any to cover server side rendering
+  const {pathname} = (window as any)?.location || {};
```
QuintonC, line 118: Nice attention to detail! 🙌

QuintonC (approved):
Looks good! Nice work 👏

PR #81, opened 2023-03-10
:wrench: Configure TailwindCSS monorepo config

QuintonC:

Initializes the TailwindCSS migration for our packages.

This is the first PR which really just adds TailwindCSS to the different packages. It doesn't convert any of the components yet.

While this might not be _everything_ we need for the configuration, I'm hopeful that it's a mighty first step in the right direction.

## ✅ Checklist

- [x] ~Tested on mobile~ N/A
- [x] ~Tested on multiple browsers~ N/A
- [x] ~Tested for accessibility~ N/A
- [x] ~Includes unit tests~ N/A
- [x] ~Updated relevant documentation for the changes (if necessary)~ N/A


package.json
```diff
@@ -66,15 +66,20 @@
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
     "@vitejs/plugin-react": "^2.2.0",
+    "autoprefixer": "^10.4.14",
     "babel-plugin-react-require": "^3.1.3",
+    "concurrently": "^7.6.0",
```
QuintonC, line 71: `concurrently` is used to run the Tailwind CLI tasks for building stylesheets in parallel along the tsup dev processes.

packages/connect-wallet/.prettierrc.cjs
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  ...require('@shopify/prettier-config'),
+  plugins: [require('prettier-plugin-tailwindcss')],
```
QuintonC, line 3: The [prettier-plugin-tailwindcss](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) plugin will help us keep our classNames in a [uniform order](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted)

packages/connect-wallet/package.json
```diff
@@ -24,10 +24,19 @@
...
   "node": ">=14.16",
   "types": "./dist/index.d.ts",
-  "exports": "./dist/index.js",
+  "exports": {
+    ".": {
+      "default": "./dist/index.js"
+    },
+    "./styles.css": {
+      "import": "./dist/styles.css"
+    }
```
QuintonC, line 33: This ensures that the styles are exported out of the package and can be imported

```ts
import '@shopify/connect-wallet/style.css';
```

packages/connect-wallet/package.json
```diff
@@ -24,10 +24,19 @@
...
+  },
   "scripts": {
-    "build": "tsup",
-    "dev": "DEV=true tsup --watch",
+    "build": "tsup && tailwindcss -i ./src/styles.css -o ./dist/styles.css",
+    "dev": "concurrently \"yarn run dev:*\"",
+    "dev:tailwind": "tailwindcss -i ./src/styles.css -o ./dist/styles.css --watch",
```
QuintonC, line 38: Responsible for building the style.css file.

`-i` is the input
`-o` is the output
`--watch` ensures that any time a path within the tailwind config's `content` section is changed that the styles are rebuilt.

packages/connect-wallet/package.json
```diff
@@ -24,10 +24,19 @@
...
+    },
+    "./styles.css": {
+      "import": "./dist/styles.css"
+    }
+  },
   "scripts": {
-    "build": "tsup",
-    "dev": "DEV=true tsup --watch",
+    "build": "tsup && tailwindcss -i ./src/styles.css -o ./dist/styles.css",
+    "dev": "concurrently \"yarn run dev:*\"",
```
QuintonC, line 37: Runs the tsup and tailwind commands in parallel.

packages/connect-wallet/postcss.config.cjs
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
QuintonC, line 1: This is the default config for postcss config with tailwind.

packages/connect-wallet/src/styles.css
```diff
@@ -0,0 +1,14 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+@layer base {
+  * {
+    @apply sbcc-box-border;
```
QuintonC, line 7: This makes sure that we're utilizing `box-sizing: box-border;` throughout our components.

packages/connect-wallet/src/styles.css
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  [id^='shopify-connect-wallet'] *:empty,
+  #walletconnect-wrapper *:empty {
```
QuintonC, line 11: This will ensure that anything matching an id that starts with `shopify-connect-wallet` and an empty container still renders. This is particularly helpful for the background overlays of the modal and popover components.

packages/tailwind-config/tailwind.config.js
```diff
@@ -0,0 +1,121 @@
+const defaultTheme = require('tailwindcss/defaultTheme');
+
+module.exports = {
+  // Using SBCC for now
+  // [S]hopify [B]lock[c]hain [C]omponents
```
QuintonC, line 5: Prefixes ensure that our styles don't clash with any other tailwind styles which may be present. I feel like sbcc is concise and somewhat unique, but if we feel we need to change the prefix we can!

While it's not guaranteed to prevent style conflicts, it's better than _not_ using prefixes and only using the default classes :)

packages/tailwind-config/tailwind.config.js
```diff
@@ -0,0 +1,121 @@
...
+  prefix: 'sbcc-',
+  content: [
+    '../../packages/connect-wallet/src/**/*.{ts,tsx}',
+    '../../packages/shared/src/**/*.{ts,tsx}',
+    '../../packages/tokengate/src/**/*.{ts,tsx}',
```
QuintonC, line 10: The leading up directories here are just relative to the shared configuration's location which ensures that all paths for the tailwind config match and pick up changes to files as intended.

packages/tailwind-config/tailwind.config.js
```diff
@@ -0,0 +1,121 @@
...
+  ],
+  theme: {
```
QuintonC, line 12: The following theme config matches our default theme.

I've not yet pieced out how to use the dawn theme. That will require some more reading and possibly configuration, however I'd prefer to focus on the default theme for the time being.

packages/tailwind-config/tailwind.config.js
lesliexin: Thanks for explaining this! Is there a reason why we went with SBCC over SBC?

packages/tailwind-config/tailwind.config.js
```diff
@@ -0,0 +1,120 @@
...
+    },
+    borderRadius: {
+      'button-large': '6px',
+      'button-medium': '6px',
+      'button-small': '6px',
+      'modal-desktop': '16px',
+      'modal-mobile': '32px 32px 0px 0px',
+      popover: '8px',
+      tokengate: '8px',
+      // Add the DEFAULT value -- DEFAULT is a Tailwind convention
```
lesliexin, line 44: What does this comment mean / was it meant to be resolved?

lesliexin (approved):
Thanks for taking this on! Just a couple questions out of curiosity

packages/tailwind-config/tailwind.config.js
QuintonC: > Thanks for explaining this! Is there a reason why we went with SBCC over SBC?

I'm fine with that as well :) 

I'll convert the classNames over after this PR merges to the feature branch!

PR #82, opened 2023-03-10
[analytics] Track `connect-wallet` and `tokengate` rendering

caropinzonsilva:

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.test.tsx
```diff
@@ -156,4 +157,27 @@ describe('utils', () => {
       expect(subscriberMock).toBeCalledTimes(0);
     });
   });
+
+  describe('useComponentRenderedTracking', () => {
+    const onClickEventName = 'TEST_EVENT_NAME';
```
QuintonC, line 162:
```suggestion
    const onRenderedEventName = 'TEST_EVENT_NAME';
```

packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
```diff
@@ -1,5 +1,6 @@
 export const eventNames = {
   TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+  CONNECT_WALLET_COMPONENT_RENDERED: 'CONNECT_WALLET_COMPONENT_RENDERED',
```
QuintonC, line 3:
```suggestion
  CONNECT_WALLET_PROVIDER_MOUNTED: 'CONNECT_WALLET_PROVIDER_MOUNTED',
```

My thought here is that this is currently tracking when the `ConnectWalletProvider` renders. I might be wrong, but I'm thinking we might want more granular analytics here, such as:
- CONNECT_BUTTON_RENDERED
- CONNECT_MODAL_RENDERED
- CONNECTED_BUTTON_RENDERED

Probably splitting hairs here, but I thought it was something worth pointing out. What are your thoughts on this?

packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.test.tsx
```diff
@@ -156,4 +157,27 @@ describe('utils', () => {
...
+    const Element = () => {
+      useComponentRenderedTracking('TEST_EVENT_NAME');
+      return <div />;
+    };
```
QuintonC, line 166: Interesting approach! Just for learning purposes, is there a reason you chose this approach instead of using `renderHook` [testing library](https://testing-library.com/docs/react-testing-library/api#renderhook)?

packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.ts
```diff
@@ -106,3 +106,13 @@ export const useEventWithTracking = ({
...
   );
+
+/**
+ * @param {string} eventName An event name indicating which event to publish to subscribers.
+ * @returns {void}
+ */
+export const useComponentRenderedTracking = (eventName: string) => {
+  useEffect(() => {
+    publishEvent(eventName);
+  }, [eventName]);
```
QuintonC, line 117: I'm wondering if we can omit the dep array since (_I think_) the goal is to only run this on mount. What are your thoughts?

packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
caropinzonsilva: Makes sense. For now, we will only track the provider but it makes sense to have the event name be more explicit about this

packages/blockchain-components/src/utils/analytics/ClientAnalytics/utils.test.tsx
caropinzonsilva: I prefer this approach so because it was more like an integration test, rather than a unit test. With the `renderHook` we would not really be rendering the component. Additionally, with the `AnalyticsListenerTestHelper` we test both that the event is published and that the subscriber captures it.

packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
caropinzonsilva: Addressed here https://github.com/Shopify/blockchain-components/pull/82/commits/c09c8d2d16fa1aaddc56c69de1382530dc766560

jamiely (approved):
:+1: I did not tophat

PR #83, opened 2023-03-10
💄 Makes use of Tailwind in shared Button

QuintonC:
⚠️ Fixes: 

Converts the `<Button />` component in the internal `shared` package to Tailwind.

## 🕹️ Demonstration

Note in the image below for the `Primary` variant on the button that the label has an incorrect color.

Since this pull request is only focused on the Button itself, I'll update the Text component in a follow-up PR.

| Component | Image |
| - | - |
| ConnectButton |  SCR-20230310-ku7 |
| Disconnect | SCR-20230310-kup |

## ✅ Checklist

- [x] ~Tested on mobile~ N/A
- [x] ~Tested on multiple browsers~ N/A
- [x] ~Tested for accessibility~ N/A
- [x] ~Includes unit tests~ N/A
- [x] ~Updated relevant documentation for the changes (if necessary)~ N/A


packages/shared/src/components/Button/Button.tsx
```diff
@@ -2,21 +2,49 @@ import {useEventWithTracking} from '@shopify/blockchain-components';
...
+const MD_CSS: ClassName = 'sbcc-rounded-button-medium sbcc-p-button-medium';
+const SM_CSS: ClassName = 'sbcc-rounded-button-small sbcc-p-button-small';
+
+// Variants
+const PRIMARY_CSS: ClassName =
+  'sbcc-bg-button-primary sbcc-border-button-primary sbcc-text-button-primary hover:sbcc-bg-button-primary-hover';
+const SECONDARY_CSS: ClassName =
+  'sbcc-bg-button-secondary sbcc-border-button-secondary sbcc-text-button-secondary hover:sbcc-bg-button-secondary-hover';
+const DISABLED_CSS: ClassName =
+  'sbcc-bg-button-disabled sbcc-border-button-disabled sbcc-text-button-disabled';
```
QuintonC, line 20: Note: we don't have to worry about the `cursor: default` style here because [Tailwind preflight](https://github.com/tailwindlabs/tailwindcss/blob/master/src/css/preflight.css#L334-L336) takes care of the button cursors.

PR #84, opened 2023-03-10
💄 Makes use of Tailwind in shared IconButton

QuintonC:

- Converts the `<IconButton />` component in the internal `shared` package to Tailwind
- Removes the relevant `style.ts` file for the component.

## 🕹️ Demonstration

The IconButton is only used in the Modal for the `Back` button, the `What are wallets` button, and the `Close` button
| What are wallets + Close | Back + Close |
| - | - |
| SCR-20230310-mot | SCR-20230310-lyq |


apps/playground/src/App.tsx
```diff
@@ -5,7 +5,6 @@ import {
   adaptUnlockingTokens,
 } from '@shopify/tokengate';
 
-import './DawnVariables.css';
```
QuintonC, line 8: DawnVariables contained `:root` values which overrides values for our style configuration. Specifically, there was a font-size: 10px value which was conflicting with the way that Tailwind height and width styles were applied as the basis for `rem` was set to `10px` rather than `16px`

PR #89, opened 2023-03-13
[Tailwind] shared/Spinner

QuintonC:

Converts the `<Spinner />` component in the internal `shared` package to Tailwind.

## 🕹️ Demonstration

image

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A


packages/tailwind-config/tailwind.config.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
   ],
   theme: {
     extend: {},
+    animation: {
+      spin: 'spin 500ms linear infinite',
+    },
```
QuintonC, line 16: The default `animate-spin` class has a **1 second** duration, but we use 500ms, so this just updates the spin animation to use `500ms` instead of `1s`

[More here](https://tailwindcss.com/docs/animation)

PR #93, opened 2023-03-13
[Tailwind] tokengate/Error

QuintonC:

Makes use of Tailwind in `@shopify-tokengate/Error` component.

I had to modify `shared/Text` to add a record for colors. The reason behind this is that without this, Tailwind was not picking up the dynamic `className` for `sbc-text-${color}`. This led to `sbc-text-critical` not being defined or compiled.  There are two approaches to resolving this, and I chose one that resulted in more code up front, but less compiled code.

- [Safelisting classes](https://tailwindcss.com/docs/content-configuration#safelisting-classes)
  - I **did not** choose this path because it resulted in additional classes being compiled during the build process. When safelisting with a pattern for the text-colors, I received classes for all of the opacity ranges of the text colors.
- Explicitly defining the color classes that will be used (and should be compiled).

## 🕹️ Demonstration

image

[📖  Storybook](https://639b1f308693132693d9b82c-mhvxnbcaaz.chromatic.com/?path=/story/tokengate-exclusive-unlocked--order-limit-reached)

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


jamiely (approved):
:+1:

QuintonC:
Update -- rebased to factor in bottom padding on the card. Previous commit would have had 8px additional on the bottom of the card.

PR #103, opened 2023-03-13
[Tailwind] Remove styled components usage from `packages/tokengate`

QuintonC:

Removes all references of `styled` from the `Tokengate` package with one exception which would be the changelog.

The last remaining piece was primarily to convert the `playground` components.

## 🕹️ Demonstration

image

[📖  Storybook](https://639b1f308693132693d9b82c-urttaqwjmh.chromatic.com/?path=/story/tokengate--playground)


packages/tokengate/src/components/TokenList/index.ts
```diff
@@ -1,4 +1,2 @@
 export {TokenList} from './TokenList';
-export {TokenListImage} from './TokenListImage';
-export {TokenListSkeleton} from './TokenListSkeleton';
```
QuintonC, line 3: Neither of these are used outside of `TokenList`

packages/tokengate/src/components/TokengateRequirements/index.ts
```diff
@@ -1,2 +1 @@
 export {TokengateRequirements} from './TokengateRequirements';
-export {TokengateRequirementsSeparator} from './TokengateRequirementsSeparator';
```
QuintonC, line 2: This is only used inside of `TokengateRequirements`

jamiely (approved):
Nice, thanks!

PR #111, opened 2023-03-14
[analytics] Add clientBrowserParams utils

caropinzonsilva:

...

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/blockchain-components/src/utils/clientBrowserParameters/types.ts
```diff
@@ -0,0 +1,50 @@
+export interface ClientBrowserParameters {
+  /**
+   * Shopify unique user token: Value of `_shopify_y` cookie.
+   *
+   * Use `getClientBrowserParameters()` to collect this value.
```
QuintonC, line 5: I'm assuming the `getClientBrowserParameters` comment will be removed when we begin using it?

packages/blockchain-components/src/utils/clientBrowserParameters/index.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Utils
+ */
+export {getClientBrowserParameters} from './utils';
```
QuintonC, line 4: This looks like a comment we would place at the top-level `index.ts` for the package. Was it intended for this to be in both this `index.ts` and `src/index.ts`? places?

QuintonC (approved):
Looks good, just a couple of non-blocking questions

packages/blockchain-components/src/utils/clientBrowserParameters/types.ts
caropinzonsilva: This was a copy + paste, but I can remove it! I think what the comment is saying is that if anyone wants to use this type in the code, they need to use that function.

packages/blockchain-components/src/utils/clientBrowserParameters/index.ts
caropinzonsilva: It was! The `index` in the `clientBrowserParameters` is for when we use it in other files inside this package. The root index is for outside.

packages/blockchain-components/src/utils/clientBrowserParameters/types.ts
QuintonC: Gotcha. I'm fine with either way, I just wanted to check. Thanks!

PR #117, opened 2023-03-14
[Analytics]Add event tracking for gate context being added to cart attributes

lesliexin:

Adds the ability to track when the context is submitted to the cart attributes in the `gate-context-client` package.

Since the `gate-context-client` package requires that you're on the online store to write to the cart, we can't tophat it in the playground. 

@jamiely and I decided to take a snapshot of this PR to be able to test once the gate evaluation function is fixed through the example app or CommerceTown.

## ✅ Checklist

- [x] ~~Tested on mobile~~
- [x] Includes unit tests


packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
```diff
@@ -32,6 +28,12 @@ export const eventNames = {
   CONNECT_WALLET_ON_CONNECT_EVENT: 'CONNECT_WALLET_ON_CONNECT_EVENT',
   CONNECT_WALLET_ON_SIGN_MESSAGE_EVENT: 'CONNECT_WALLET_ON_SIGN_MESSAGE_EVENT',
   CONNECT_WALLET_ON_DISCONNECT_EVENT: 'CONNECT_WALLET_ON_DISCONNECT_EVENT',
+  TOKENGATE_COMPONENT_RENDERED: 'TOKENGATE_COMPONENT_RENDERED',
+  TOKENGATE_ON_UNLOCK_EVENT: 'TOKENGATE_ON_UNLOCK_EVENT',
+  TOKENGATE_ON_REQUIREMENTS_NOT_MET_EVENT:
+    'TOKENGATE_ON_REQUIREMENTS_NOT_MET_EVENT',
```
lesliexin, line 34: Alphabetizing

packages/blockchain-components/src/utils/analytics/ClientAnalytics/const.ts
```diff
@@ -32,6 +28,12 @@ export const eventNames = {
...
+  GATE_CONTEXT_CLIENT_ON_CONTEXT_SUBMITTED_TO_CART_ATTRIBUTES:
+    'GATE_CONTEXT_CLIENT_ON_CONTEXT_SUBMITTED_TO_CART_ATTRIBUTES',
```
caropinzonsilva, line 36: I think the event's name, for now, is ok. Although they are kind of verbose, it is better to have a long descriptive name than a short one that is hard to understand.

PR #121, opened 2023-03-14
[Tailwind] Removes all references to styled-components

QuintonC:

The conversion to Tailwind is **complete**.

Therefore, this PR removes **all references** to **anything** which used styled-components. You should primarily be looking for:
- ✅ Green CI
- ✅ A functioning storybook with the **default theme** present in all components. I will add examples on how developers can theme these components in an upcoming PR (this is still merging into the feature branch).

Files removed (with an explanation where necessary)

| File | Explanation |
| - | - |
| `.storybook/decorators/index.ts` | See next row. |
| `.storybook/decorators/withThemeProvider.tsx` | Since we no longer have styled-components, we can no longer style our stories using predefined themes. At least, not for now. However, we can potentially implement themes in the future if we want to add theming to Storybook, however, I think the mindset we're wanting to shift towards is to ship a single theme and allow the components to be re-themed to fit the merchant's needs. |
| `apps/playground/src/style.ts` | This was only a simple style file which contained CSS / styling for the preview in the local testing environment. |
| `apps/playground/styled.d.ts` | A declaration file which allowed us to inherit typings for the `DefaultTheme` type, which we no longer need with styled-components being removed. |
| `packages/connect-wallet/styled.d.ts` | A declaration file which allowed us to inherit typings for the `DefaultTheme` type, which we no longer need with styled-components being removed. |
| `packages/shared/src/components/Button/style.ts` | A styled components style file. This was used to export `ButtonWrapper` which was used up until the end of the `Modal` conversions in the `connect-wallet` package. |
| `packages/shared/src/providers/RootProvider.tsx` | This is what our components were using to receive the `theme` prop and pass it into context for our components to use. |
| `packages/shared/src/styles/sharedStyles.ts` | This was the global styles which we used for overriding `*:empty` previously. Tailwind handles these for us (and we've added classes + ids which will handle the non-empty div display for us) |
| `packages/shared/src/styles/styled.ts` | This was the styled-components hack put in place to support styled-components in webpack-based frameworks. |
| `packages/shared/src/tests/utils.ts` | This was only a simple react testing library render export which used the `RootProvider` component which only used a theme. We can just use the `render` export from react testing library instead. |
| `packages/shared/src/themes/dawn.ts` | We no longer export themes since they are for styled-components. |
| `packages/shared/src/themes/default.ts` | We no longer export themes since they are for styled-components. |
| `packages/shared/src/themes/index.ts` | We no longer export themes since they are for styled-components. |
| `packages/shared/src/types/theme.ts` | We no longer export themes since they are for styled-components. |
| `packages/shared/styled.d.ts` | A declaration file which allowed us to inherit typings for the `DefaultTheme` type, which we no longer need with styled-components being removed. |

[📖 Storybook](https://639b1f308693132693d9b82c-mshncozxsp.chromatic.com/?path=/story/connect-wallet--playground&globals=backgrounds.value:!hex(F8F8F8))


jamiely (approved):
I tophatted in Storybook and confirmed components appeared. One thing I was wondering about is it seems there is inconsistent font sizing between certain button labels. Is this expected?

image

QuintonC:
> I tophatted in Storybook and confirmed components appeared. One thing I was wondering about is it seems there is inconsistent font sizing between certain button labels. Is this expected?
> 
> 
> 
> image
> 
> 

Yeah, but that's only because the buttons used in storybook for the Tokengate component are not the actual connect wallet buttons. They're only mock buttons placed inside of the stories.

PR #122, opened 2023-03-15
[Tailwind] Updates breakpoints to tailwind sizes

QuintonC:

Updates `breakpoints` to reflect the [breakpoints defined by Tailwind](https://tailwindcss.com/docs/responsive-design) to ensure that our CSS matches up with our `useMediaQuery` hooks (used in `Popover`).

You can use Storybook for testing popovers on different mobile devices, however, the thing you're mostly looking for here is green CI.

[📖 Storybook](https://639b1f308693132693d9b82c-zvgvrlexlx.chromatic.com/)

### DIFF
--- a/packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
+++ b/packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
@@ -3,7 +3,7 @@ import {
   useComponentRenderedTracking,
 } from '@shopify/blockchain-components';
 import {MotionConfig} from 'framer-motion';
-import {FC, useMemo} from 'react';
+import {FC, PropsWithChildren, useMemo} from 'react';
 import {Provider} from 'react-redux';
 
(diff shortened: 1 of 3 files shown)


packages/connect-wallet/src/providers/ConnectWalletProvider/provider.tsx
```diff
@@ -14,14 +14,14 @@ import store from '../../store/configureStore';
 import {ConnectWalletContext, ConnectWalletProviderValue} from './context';
 import {ProviderProps} from './types';
 
-export const ConnectWalletProvider: FC<ProviderProps> = ({
+export const ConnectWalletProvider: FC<PropsWithChildren<ProviderProps>> = ({
```
QuintonC, line 17: Note: `PropsWithChildren` was re-added here because we are not longer utilizing the `ThemeProvider` / `RootProvider` since it was removed in #121

PR #126, opened 2023-03-15
[analytics] Use window.ShopifyAnalytics to get service name

caropinzonsilva:

...

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/blockchain-components/src/utils/shopify/utils.test.ts
```diff
@@ -1,84 +1,149 @@
...
 
-    describe('edge cases', () => {
-      it('returns OTHER for unknown services', () => {
-        expect(getShopifyService('/')).toBe(shopifyServices.OTHER.name);
-        expect(getShopifyService('/product-')).toBe(shopifyServices.OTHER.name);
-        expect(getShopifyService('/otherPathname')).toBe(
-          shopifyServices.OTHER.name,
-        );
+    describe('using pathname', () => {
+      describe("real shop's URLs", () => {
```
caropinzonsilva, line 63: The diff is making it look like all the tests are new, but I just wrapped the previous tests in a `describe('using pathname')`.

packages/blockchain-components/src/utils/shopify/utils.test.ts
```diff
@@ -1,84 +1,149 @@
...
+          expect(getShopifyServiceName(realShopUrls.commerceTown.page)).toBe(
+            shopifyServices.PAGE.name,
+          );
+        });
+      });
+
+      describe('edge cases', () => {
+        it('return Home for / pathname', () => {
+          expect(getShopifyServiceName('/')).toBe(shopifyServices.HOME.name);
+        });
```
caropinzonsilva, line 136: This is also a new addition. If the pathname is `/`, we can assume that the service is the `HOME`. It might not work with some edge cases, but they will be easy to identify.

jamiely (approved):
Looks good, thanks!

PR #130, opened 2023-03-15
Add wallet extensibility changes to CRA example app

lesliexin:

Refactors the CRA example app to use the new `buildConnectors` function and to pass `connectors` to the `ConnectWalletProvider`. The changes are the same that exist in our [playground](https://github.com/Shopify/blockchain-components/tree/main/apps/playground). 

## 🕹️ Demonstration

https://user-images.githubusercontent.com/18248358/225454608-e65bcb60-20e8-4cb6-9cd2-a4a1f4a12783.mov

- Enter the `examples/cra` directory 
- Run `yarn start`
- Verify that the app runs and that you can connect your wallet

## ✅ Checklist

- [x] ~~Tested on mobile~~
- [x] ~~Includes unit tests~~

### DIFF
--- a/packages/connect-wallet/src/types/connector.ts
+++ b/packages/connect-wallet/src/types/connector.ts
@@ -1,5 +1,4 @@
-import {Chain} from '@wagmi/core';
-import {Connector as WagmiConnector} from 'wagmi';
+import {Chain, Connector as WagmiConnector} from 'wagmi';
 import {InjectedConnector} from 'wagmi/connectors/injected';
 import {WalletConnectConnector} from 'wagmi/connectors/walletConnect';
 
--- a/examples/cra/package.json
+++ b/examples/cra/package.json
@@ -23,10 +23,12 @@
     "tsconfig": "file:../../packages/tsconfig"
   },
   "resolutions": {
+    "@shopify/blockchain-components": "file:../../packages/blockchain-components",
     "@shopify/gate-context-client": "file:../../packages/gate-context-client"
   },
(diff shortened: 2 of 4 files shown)


packages/connect-wallet/src/types/connector.ts
```diff
@@ -1,4 +1,4 @@
-import {Chain} from '@wagmi/core';
+import {Chain} from 'wagmi';
```
lesliexin, line 1: Running the CRA was throwing the following error:

image

It seemed to be the: `type Chains are incompatible` that was causing the error. 

Eventually @QuintonC helped figure out that it was where we imported type `Chains` in the `connect-wallet` package that was causing the type error. 

image

We're still not sure why it errored in the CRA but not with Vite or Next.

packages/connect-wallet/src/types/connector.ts
```diff
@@ -1,4 +1,4 @@
...
 import {Connector as WagmiConnector} from 'wagmi';
```
QuintonC, line 2: Thanks for adding context. It looks like you have a lint violation due to importing the same package twice (duplicate imports).

QuintonC (approved):
👍 LGTM. Approving with the assumption that the lint violation will be resolved prior to merging.

packages/connect-wallet/src/types/connector.ts
lesliexin: Yup, just fixed it!

PR #133, opened 2023-03-16
Chore/tailwind theming support

QuintonC:

- Adds custom theme support for components via CSS variables as suggested by @benjaminsehl (thank you!)
- Includes an example in the `/examples` directory, nested under `vite-custom-theme`.

Some good takeaways from this:
- Since this uses CSS variables, any time the variable changes (for example if the color mode changes), the components will update with the new colors.
- Instead of being required to provide a theme to either `ConnectWalletProvider` or `Tokengate` components, the developer can adjust as many or as few variables as they want to obtain their desired UI.

## 🕹️ Demonstration

https://user-images.githubusercontent.com/4250423/225481861-3c2a0286-0d05-4f84-b0e6-11efd4a67da5.mov

1. Pull this branch
2. Ensure you've run `yarn install` at the repository root directory
3. Navigate to the Vite example application
4. Ensure you do not have `node modules` or a `yarn.lock`
5. Run `yarn install` in the Vite example
6. Run `yarn dev`

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [ ] POST-MERGE:Updated relevant documentation for the changes (if necessary)


packages/connect-wallet/src/components/ConnectButton/Popover.tsx
```diff
@@ -100,14 +100,13 @@ export const Popover = ({mobile, onDismiss, visible}: PopoverProps) => {
               <ConnectorIcon id={connectorId} size="lg" />
 
               <button
-                className="sbc-flex sbc-cursor-pointer sbc-items-center sbc-gap-x-3 sbc-rounded-full sbc-bg-address-chip sbc-py-2 sbc-px-3 sbc-transition-colors sbc-border-none hover:sbc-bg-button-secondary-hover"
+                className="sbc-flex sbc-cursor-pointer sbc-items-center sbc-gap-x-3 sbc-rounded-full sbc-bg-address-chip sbc-py-2 sbc-px-3 sbc-text-address-chip sbc-transition-colors sbc-border-none hover:sbc-bg-address-chip-hover"
```
QuintonC, line 103: I added `bg-address-chip` and its `-hover` counterpart as theme values in our tailwind config so developers could customize the address chip within the popover independently of the buttons.

packages/shared/src/components/Spinner/Spinner.tsx
```diff
@@ -1,12 +1,15 @@
...
+    <div className="sbc-mx-auto sbc-my-0 sbc-h-5 sbc-w-5 sbc-text-primary">
       <svg
         className="sbc-animate-spin"
         viewBox="0 0 20 20"
         xmlns="http://www.w3.org/2000/svg"
       >
-        <path d="M7.229 1.173a9.25 9.25 0 1011.655 11.412 1.25 1.25 0 10-2.4-.698 6.75 6.75 0 11-8.506-8.329 1.25 1.25 0 10-.75-2.385z" />
+        <path
+          d="M7.229 1.173a9.25 9.25 0 1011.655 11.412 1.25 1.25 0 10-2.4-.698 6.75 6.75 0 11-8.506-8.329 1.25 1.25 0 10-.75-2.385z"
+          fill="currentColor"
```
QuintonC, line 11: Makes the spinner inherit the current primary text color.

packages/tailwind-config/tailwind.config.js
```diff
@@ -16,113 +18,114 @@ module.exports = {
...
-      'body-sm': defaultTheme.fontSize.xs,
-      'heading-lg': defaultTheme.fontSize['2xl'],
-      'heading-md': defaultTheme.fontSize.xl,
-      'heading-sm': defaultTheme.fontSize.base,
+      'body-lg': 'var(--sbc-font-size-body-lg, 1rem)',
+      'body-md': 'var(--sbc-font-size-body-md, 0.875rem)',
+      'body-sm': 'var(--sbc-font-size-body-sm, 0.75rem)',
+      'heading-lg': 'var(--sbc-font-size-heading-lg, 1.5rem)',
+      'heading-md': 'var(--sbc-font-size-heading-md, 1.25rem)',
+      'heading-sm': 'var(--sbc-font-size-heading-sm, 1rem)',
```
QuintonC, line 101: You'll notice we're not referencing `defaultTheme` here anymore. That's because defaultTheme contained multiple values. One for the `font-size` and one for the `line-height`. When trying to retain the use of `defaultTheme` here, it would cause the CSS to break as it would attempt to add an `Object` into the css value.

jamiely (approved):
Nice example, thanks!

examples/vite-custom-theme/src/index.css
```diff
@@ -0,0 +1,137 @@
...
+  font-weight: 500;
+  font-family: inherit;
+  background-color: #1a1a1a;
+  cursor: pointer;
+  transition: border-color 0.25s;
+}
+/* button:hover {
+  border-color: #646cff;
+} */
+button:focus,
```
caropinzonsilva, line 87: I got a little bit lost with these CSS rules. Are they also part of the custom theme?

caropinzonsilva (approved):
Looks good! Just a comment about some CSS rules that looked unrelated while doing the review

PR #136, opened 2023-03-16
[Tailwind] Adds stylesheet to instructions for connect-wallet

QuintonC:

Adds the stylesheet for `connect-wallet` to the README for the package. The README for tokengate points to the dev docs, otherwise I would've added that one as well.

### DIFF
--- a/packages/connect-wallet/README.md
+++ b/packages/connect-wallet/README.md
@@ -63,6 +63,7 @@ Let's begin using the configured client and chains. In your app's entry point, (
 
 ```tsx
 import {ConnectWalletProvider} from '@shopify/connect-wallet';
+import '@shopify/connect-wallet/styles.css';
 import {WagmiConfig} from 'wagmi';
 
 import {chains, client, connectors} from './connect-wallet-config'


lesliexin (commented):
The description makes me think we should be consistent in the READMEs between the two packages. 

I'm leaning towards linking to the external docs like Tokengate, just so we have one central source of truth and less disparate things to manage in the future. What do you think?

QuintonC:
> The description makes me think we should be consistent in the READMEs between the two packages.
> 
> I'm leaning towards linking to the external docs like Tokengate, just so we have one central source of truth and less disparate things to manage in the future. What do you think?

@lesliexin IMO I can think of many reasons to keep documentation (at least the very basics of it) inside of a README. One in particular is that it is the closest thing to the codebase and having at the very least some basic information on getting started is good to have in the event that a developer opens the package's `README.md` from their `node_modules` while offline or just because they don't feel the need to navigate to an external documentation site.

With that being said, I'm not going to push back on folks saying we should remove it, but I think if that is the case then it should be done _outside_ of the scope of the Tailwind changes. This addition is only for punctuality and not missing any details within this copy of installation instructions.

PR #138, opened 2023-03-17
:truck: Moves the Popover to the shared folder

QuintonC:
⚠️ Part of: #75

PR #75 was beginning to grow a bit, so I felt it was best to move the Popover to its separate PR.

I made use of the shared component in the `@shopify/connect-wallet` package's `ConnectedButton` to ensure that it was functioning the same as it did before.

You can use the connect-wallet playground in Storybook to test the popover when your wallet is connected.

[📖  Storybook](https://639b1f308693132693d9b82c-kojgyluzku.chromatic.com/?path=/story/connect-wallet--playground)

## ✅ Checklist

- [x] Tested on mobile


packages/shared/src/components/Popover/variants.ts
```diff
@@ -0,0 +1,77 @@
+import {Variants} from 'framer-motion';
+
+import {EASE, MOBILE_SPRING, SPRING} from '../../constants/transitions';
+
+export const PopoverReducedMotionSmall: Variants = {
```
caropinzonsilva, line 5: I noticed we are leaving these variants in the `connect-wallet` package: https://github.com/Shopify/blockchain-components/blob/main/packages/connect-wallet/src/components/ConnectButton/variants.ts#L5

Do we still need them there? Or could they be removed?

caropinzonsilva (approved):
Code looks good and I 🎩 ed in Storybook.

I think there are some variants and constants that we can remove from the `connect-wallet` package (mentioned in my comments)

packages/shared/src/components/Popover/variants.ts
QuintonC: They can be removed, but I'll need to update the [Modal variants](https://github.com/Shopify/blockchain-components/blob/main/packages/connect-wallet/src/components/Modal/variants.ts#L3) as well. I thought about doing that change in this PR but I was wary since I didn't want the PR to grow _too_ much. 

I'm assuming you're okay with me doing this though, so I'll do that!

PR #139, opened 2023-03-17
:bug: Adds polyfills to storybook

QuintonC:

Adds polyfills to Storybook to ensure that QR codes properly display.

## 🕹️ Demonstration

| Before | After |
| - | - |
| Screenshot 2023-03-17 at 12 40 16 PM | Screenshot 2023-03-17 at 12 42 41 PM |

[📖  Storybook](https://639b1f308693132693d9b82c-gucxkfslez.chromatic.com/?path=/story/connect-wallet--playground)

### DIFF
--- a/examples/vite-custom-theme/vite.config.ts
+++ b/examples/vite-custom-theme/vite.config.ts
@@ -1,6 +1,6 @@
 import {defineConfig} from 'vite';
 import react from '@vitejs/plugin-react';
-import nodePolyfills from 'vite-plugin-node-stdlib-browser';
+import {nodePolyfills} from 'vite-plugin-node-polyfills';
 
 // https://vitejs.dev/config/
 export default defineConfig({
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const react = require('@vitejs/plugin-react');
+const {nodePolyfills} = require('vite-plugin-node-polyfills');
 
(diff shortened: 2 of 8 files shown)


examples/vite-custom-theme/vite.config.ts
```diff
@@ -1,6 +1,6 @@
 import {defineConfig} from 'vite';
 import react from '@vitejs/plugin-react';
-import nodePolyfills from 'vite-plugin-node-stdlib-browser';
+import {nodePolyfills} from 'vite-plugin-node-polyfills';
```
caropinzonsilva, line 3: Could we have used the old `nodePolyfills` in Storybook? I'm curious about what is the difference between them.

caropinzonsilva (approved):
Code looks good!

examples/vite-custom-theme/vite.config.ts
QuintonC: > Could we have used the old `nodePolyfills` in Storybook? I'm curious about what is the difference between them.

We could have, yes. The main difference is that the new has type declarations (it's also the same one we use in the [example app](https://github.com/Shopify/tokengating-example-app/blob/main/extensions/tokengate-src/package.json#L22).

caropinzonsilva: Makes sense!

PR #140, opened 2023-03-17
[delegate-cash] Add redux logic for supporting delegate.cash delegations

caropinzonsilva:

I'll be doing this in different PRs based on the original PR: https://github.com/Shopify/blockchain-components/pull/13

This PR adds the redux logic for supporting delegations.

## 🕹️ Demonstration
17-31-ad9o6-cr6zz

Can be 🎩 in Replit: https://replit.com/@caropinzonsilva/FamousNonstopAudit
For some reason, the QR codes are not working, so you would need to use Metamask.

- Connect your hot wallet with delegation.
- In the console.log, check the latest redux state
- You should see the `delegationsWalletAddresses` in both the connected wallets and active wallet state:
17-28-1finm-why0h

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/packages/connect-wallet/src/hooks/useMiddleware.ts
+++ b/packages/connect-wallet/src/hooks/useMiddleware.ts
@@ -6,6 +6,7 @@ import {
   addWallet,
   attributeOrder,
   fetchEns,
+  fetchDelegations,
   setActiveWallet,
   setPendingWallet,
 } from '../slices/walletSlice';
@@ -142,6 +143,7 @@ export const useMiddleware = ({
           fetchEns({address: wallet.address, chain: {...rest}, provider}),
         );
       }
+      state.dispatch(fetchDelegations(wallet.address));
     });
 
     return dispatch(listener);
--- a/packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
+++ b/packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
@@ -1,5 +1,6 @@
 import {createAsyncThunk} from '@reduxjs/toolkit';
 import {DelegateCash} from 'delegatecash';
+import {Address} from 'wagmi';
 
 export const fetchDelegations = createAsyncThunk(
   'wallet/fetchDelegations',
@@ -17,14 +18,14 @@ export const fetchDelegations = createAsyncThunk(
     );
 
     // Get only the vault wallet addresses
-    const delegationsWalletAddresses = delegationsFilteredByAllType.map(
-      (delegation) => delegation.vault,
+    const vaults = delegationsFilteredByAllType.map(
+      (delegation) => delegation.vault as Address,
     );
 
-    if (delegationsWalletAddresses.length) {
+    if (vaults.length) {
       return thunkApi.fulfillWithValue({
         address: walletAddress,
-        delegationsWalletAddresses,
+        vaults,
       });
     }
   },
--- a/packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
+++ b/packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
@@ -1,0 +1,31 @@
+import {createAsyncThunk} from '@reduxjs/toolkit';
+import {DelegateCash} from 'delegatecash';
+
+export const fetchDelegations = createAsyncThunk(
+  'wallet/fetchDelegations',
+  async (walletAddress: string, thunkApi) => {
+    const delegateCash = new DelegateCash();
+    const delegations = await delegateCash.getDelegationsByDelegate(
+      walletAddress,
+    );
+
+    // Delegation type can be 'ALL', 'CONTRACT', 'TOKEN', or 'NONE'.
+    // For delegation type 'ALL', the vault wallet delegates all actions to the delegate wallet.
+    // We filter by 'ALL' since we only want to handle delegations for the entire wallet.
+    const delegationsFilteredByAllType = delegations.filter(
+      (delegation) => delegation.type === 'ALL',
+    );
+
+    // Get only the vault wallet addresses
+    const delegationsWalletAddresses = delegationsFilteredByAllType.map(
+      (delegation) => delegation.vault,
+    );
+
+    if (delegationsWalletAddresses.length) {
+      return thunkApi.fulfillWithValue({
+        address: walletAddress,
+        delegationsWalletAddresses,
+      });
+    }
+  },
+);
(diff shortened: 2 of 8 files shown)


packages/connect-wallet/src/hooks/useMiddleware.ts
```diff
@@ -142,6 +143,7 @@ export const useMiddleware = ({
           fetchEns({address: wallet.address, chain: {...rest}, provider}),
         );
       }
+      state.dispatch(fetchDelegations(wallet.address));
```
caropinzonsilva, line 146: In the original PR, this logic was inside a `useEffect` in the `ModalProvider`: https://github.com/Shopify/blockchain-components/pull/13/files#diff-3144bd5e85f5586dd3d8e655b3b67cbd4552ca40c117527f4f38b048839f40f5R333

I moved it to the same place we fetch Ens since it makes sense that they are called at the same time (after a wallet connect).

packages/connect-wallet/src/slices/walletSlice/walletSlice.ts
```diff
@@ -8,6 +8,7 @@ import {SignatureResponse, Wallet} from '../../types/wallet';
 import {ConnectWalletError} from '../../utils/error';
 
 import {fetchEns} from './fetchEns';
+import {fetchDelegations} from './delegateCash';
```
caropinzonsilva, line 11: Original code: https://github.com/Shopify/blockchain-components/pull/13/files#diff-202b9cfc0c48f5811e30984d5d2d2ca011515a37f5fc144657766f4ae9721f50R142

packages/connect-wallet/src/hooks/useMiddleware.ts
QuintonC: Very nice!

packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
```diff
@@ -0,0 +1,31 @@
+import {createAsyncThunk} from '@reduxjs/toolkit';
+import {DelegateCash} from 'delegatecash';
+
+export const fetchDelegations = createAsyncThunk(
```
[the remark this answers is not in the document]
QuintonC: Awesome, thank you! I think this name is better and as you mentioned follows the delegate cash naming convention which I'd prefer to follow so we don't get confused in the future.

packages/connect-wallet/src/slices/walletSlice/walletSlice.ts
```diff
@@ -234,6 +235,27 @@ export const walletSlice = createSlice({
 
       throw new ConnectWalletError(errorMessage);
     });
+    builder.addCase(fetchDelegations.fulfilled, (state, action) => {
+      if (action.payload) {
+        const {address, delegationsWalletAddresses} = action.payload;
+
+        const connectedWallet = state.connectedWallets.find(
+          (wallet) => wallet.address === address,
+        );
```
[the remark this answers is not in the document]
QuintonC: Nice! This is a great improvement.

packages/connect-wallet/src/types/wallet.ts
```diff
@@ -50,4 +50,9 @@ export interface Wallet extends ConnectedWallet {
    * ISO datetime string in which the wallet was verified.
    */
   signedOn?: string;
+  /**
+   * A list of "vault" proxy wallet addresses delegations that have delegated ownership to this wallet.
+   * You can treat these addresses as being controlled by the same entity as this wallet.
+   */
+  delegationsWalletAddresses?: string[];
```
QuintonC, line 57: Three questions: 
1. Can we move this to the `ConnectedWallet` interface?
2. Can we rename this to `delegations`? `delegationsWalletAddresses` seems a bit verbose. By nature they are already wallet addresses, so I think just `delegations` would be fine.
3. Can we set the type of `delegations` to `Address[]`?

I had originally intended for the `Wallet` type to be comprised of a union of types that were separated by their concerns. For example, signature data, core wallet data, and now delegations.

Example:

```ts
export interface ConnectedWallet { // Side note, I'd be fine with naming this to WalletBaseData or something along those lines.
  /**
   * The public address of the connected wallet.
   */
  address: Address;
  /**
   * ISO datetime string in which this address was connected.
   */
  connectedAt?: string;
  /**
   * The connector id associated with how this address was connected.
   * This is particularly helpful for when a user wants to disconnect
   * their wallet and informing the user of how it was connected so
   * they can disconnect externally.
   */
  connectorId?: string;
  /**
   * The connector name associated with how this address was connected.
   */
  connectorName?: string;
  /**
   * The ENS name for the associated address.
   *
   * This will only be filled when an Alchemy or Infura provider are present.  // This can actually be removed
   */
  displayName?: string;
}

interface WalletSignatureData {
  /**
   * The message which the wallet signed. Will be undefined
   * in the event that the wallet has not yet signed the message.
   */
  message?: string;
  /**
   * The signed message from the signature transaction. Will be
   * undefined in the event that the wallet has not yet signed the message.
   */
  signature?: string;
  /**
   * ISO datetime string in which the wallet was verified.
   */
  signedOn?: string;
}

export type Wallet = ConnectedWallet & WalletSignatureData & {
  /**
   * A list of "vault" proxy wallet addresses delegations that have delegated ownership to this wallet.
   * You can treat these addresses as being controlled by the same entity as this wallet.
   */
  delegations: Address[];
}
```

QuintonC (commented):
Really great work!! I left a couple of questions around the types in `Wallet`.

packages/connect-wallet/src/slices/walletSlice/walletSlice.ts
```diff
@@ -234,6 +235,27 @@ export const walletSlice = createSlice({
 
       throw new ConnectWalletError(errorMessage);
     });
+    builder.addCase(fetchDelegations.fulfilled, (state, action) => {
+      if (action.payload) {
+        const {address, delegationsWalletAddresses} = action.payload;
+
+        const connectedWallet = state.connectedWallets.find(
+          (wallet) => wallet.address === address,
```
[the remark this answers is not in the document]
caropinzonsilva: Thanks! Addressed here: https://github.com/Shopify/blockchain-components/pull/140/commits/64b7253841e6ca77a5c2d7be61f0b53cd5687231

packages/connect-wallet/src/types/wallet.ts
QuintonC: > Thanks! Addressed here: [64b7253](https://github.com/Shopify/blockchain-components/commit/64b7253841e6ca77a5c2d7be61f0b53cd5687231)
> 
> I moved the type to `ConnectedWallet`, change it to type `Address`, and renamed it to `vaults` to follow delegate.cash naming conventions.

Amazing -- thank you!! ⭐

PR #142, opened 2023-03-17
:fire: Removes ModalProvider and utilizes Redux

QuintonC:

During a pairing session, @caropinzonsilva and I identified a bug where the modal would not reappear if the page was refreshed during a signature request.

We suspected that the issue was likely with the useState hook for the modal's active / open state. As a test, I moved the "open" state to Redux and I found that the issue was resolved.

This led to a bigger discovery; we can actually _remove_ the provider and utilize Redux for the modal state and behavior. In doing so, we remove the use of context as a state management tool and move to an actual state management tool that we are already using.

This change adds the logic that was previously in the ModalProvider into its own Redux slice (`modalSlice`).

### Before

https://user-images.githubusercontent.com/4250423/226068041-64f0f384-044b-478a-bb5f-8d41f95bca8d.mov

### After

https://user-images.githubusercontent.com/4250423/226059278-0c47f71e-720b-4002-b0d8-304d8c8d052c.mov

[📖 Storybook](https://639b1f308693132693d9b82c-rcnocfpsdr.chromatic.com/?path=/story/connect-wallet--playground)

## ✅ Checklist

- [x] Tested on mobile
- [x] Includes unit tests

### DIFF
--- a/packages/connect-wallet/src/components/Modal/Screens/SignatureScreen.tsx
+++ b/packages/connect-wallet/src/components/Modal/Screens/SignatureScreen.tsx
@@ -4,6 +4,7 @@ import {Button, Spinner, Text} from 'shared';
 
 import {ConnectorIcon} from '../../ConnectorIcon';
 import {useAppDispatch, useAppSelector} from '../../../hooks/useAppState';
+import {useDisconnect} from '../../../hooks/useDisconnect';
 import {useSignMessage} from '../../../hooks/useSignMessage';
 import {useTranslation} from '../../../hooks/useTranslation';
 import {closeModal, setError} from '../../../slices/modalSlice';
@@ -12,18 +13,20 @@ const SignatureScreen = () => {
   const dispatch = useAppDispatch();
   const {error, signing} = useAppSelector((state) => state.modal);
   const {pendingWallet} = useAppSelector((state) => state.wallet);
+  const {disconnect} = useDisconnect();
   const {signMessage} = useSignMessage();
   const {t} = useTranslation('Screens');
 
   const handleSignMessage = useCallback(() => {
     if (!pendingWallet) {
       dispatch(closeModal());
+      disconnect();
       return;
     }
 
     dispatch(setError());
     signMessage(pendingWallet);
-  }, [dispatch, pendingWallet, signMessage]);
+  }, [disconnect, dispatch, pendingWallet, signMessage]);
 
   const isCriticalError =
     error !== undefined && error.name !== 'UserRejectedRequestError';


packages/connect-wallet/src/components/Modal/Modal.tsx
```diff
@@ -36,36 +39,53 @@ import {
...
+  const {disconnect} = useDisconnect();
   const isMounted = useIsMounted();
   const escPress = useKeyPress('Escape');
   const [ref, {height}] = useMeasure();
   const isSmall = useMediaQuery('smDown');
-  const {active, closeModal, navigation} = useModal();
   const reducedMotion = useReducedMotion();
   const {t} = useTranslation('Modal');
 
+  useMiddleware({orderAttributionMode, requireSignature});
```
QuintonC, line 57: There wasn't really a better place to place this. It can't be placed inside of the `ConnectWalletProvider` since it relies on the store, so the next best place for it is inside of the Modal which is rendered within the `ConnectWalletProvider`

packages/connect-wallet/src/components/Modal/Modal.tsx
```diff
@@ -36,36 +39,53 @@ import {
...
+    if (escPress && open) {
+      dispatch(closeModal());
+    }
+  }, [dispatch, escPress, open]);
+
+  const handleCloseModal = useCallback(() => {
+    if (!open) return;
+
+    if (route === 'Signature') {
+      disconnect(pendingWallet?.address);
```
QuintonC, line 69: The disconnect functionality relies on the use of a hook, and since the purpose of Redux is to not to dispatch side effects from an action, it is best to run the disconnect functionality here when the modal is closed.

packages/connect-wallet/src/components/Modal/Modal.tsx
```diff
@@ -36,36 +39,53 @@ import {
...
-  }, [active, closeModal, escPress]);
 
-  const handleBackdropPress = useCallback(() => {
-    if (!active) return;
+    dispatch(closeModal());
+  }, [disconnect, dispatch, open, pendingWallet?.address, route]);
+
+  const handleGoBack = useCallback(() => {
+    if (route === 'Signature') {
+      disconnect(pendingWallet?.address);
```
QuintonC, line 77: The disconnect functionality relies on the use of a hook, and since the purpose of Redux is to not to dispatch side effects from an action, it is best to run the disconnect functionality here when the user navigates back route while on the `Signature` route.

packages/connect-wallet/src/hooks/useConnect.ts
```diff
@@ -0,0 +1,56 @@
+import {useContext} from 'react';
```
QuintonC, line 1: This is mostly copied code from the deleted `ModalProvider` file.

packages/connect-wallet/src/hooks/useModalContent.ts
```diff
@@ -9,38 +9,37 @@ export const useModalScreenContent = (
   const {pendingConnector} = useAppSelector((state) => state.wallet);
   const {t} = useTranslation('UseModalContent');
 
-  const screenContent: {
-    [key in keyof typeof ConnectionState]: {body: string; title: string};
```
QuintonC, line 13: `ConnectionState` changed from an enum so this was no longer valid and utilizing a Record is preferred over an object such as this.

packages/connect-wallet/src/hooks/useSignMessage.ts
```diff
@@ -1,16 +1,33 @@
...
+  validatePendingWallet,
+} from '../slices/walletSlice';
 import {SignatureResponse, Wallet} from '../types/wallet';
 import {ConnectWalletError} from '../utils/error';
 
-export function useSyncSignMessage() {
-  const {chains, statementGenerator} = useContext(ConnectWalletContext);
+import {useAppDispatch, useAppSelector} from './useAppState';
+
+export function useSignMessage() {
```
QuintonC, line 24: `useSyncSignMessage` felt improperly named, so I renamed it to `useSignMessage`

packages/connect-wallet/src/hooks/useSignMessage.ts
```diff
@@ -59,37 +76,90 @@ export function useSyncSignMessage() {
...
           signature,
         };
-      } catch (error) {
+      } catch (error: any) {
+        /**
+         * Set the error in state, resulting in an updated UI state for
+         * the signature modal. The user can attempt to sign the message
+         * with the correct wallet again.
+         */
+        dispatch(setError({message: error.message, name: error.name}));
```
QuintonC, line 153: Since it is a noop to store non-serializable data into a Redux store, we break the error down and get the message and name from it (these are the only two pieces we use in our code).

packages/connect-wallet/src/middleware/loggerMiddleware.test.ts
```diff
@@ -82,18 +82,24 @@ describe('loggerMiddleware', () => {
 
       // Test the previous state
       expect(spyLog.mock.calls[1][0]).toContain('previous state:');
-      expect(spyLog.mock.calls[1][2]).toEqual(expectedPrevState);
+      expect(spyLog.mock.calls[1][2]).toEqual(
+        expect.objectContaining(expectedPrevState),
```
QuintonC, line 86: The store now consists of two slices, so it will **contain** the previous state we are testing here.

packages/connect-wallet/src/store/combineReducers.ts
```diff
@@ -16,5 +20,6 @@ const walletPersist = {
 };
 
 export const rootReducer = combineReducers<AppState>({
+  modal: modalSlice.reducer,
```
QuintonC, line 23: the `modalSlice` does not persist any data.

jamiely (approved):
LGTM.

I noticed while tophatting that when I refresh the page during a signature request, the page opens in the same state, and seems to perform another request. Is that expected?

packages/connect-wallet/src/hooks/useSignMessage.ts
```diff
@@ -59,37 +76,90 @@ export function useSyncSignMessage() {
...
+        // Publish the signed message event
         publishEvent(eventNames.CONNECT_WALLET_ON_SIGN_MESSAGE_EVENT, {
           address,
           connector: connectorId,
         });
 
+        /**
+         * Note: We will only move past the validatePendingWallet action
+         * when the signed message is decrypted to match the address
+         * matching the pending wallet and the nonces match.
```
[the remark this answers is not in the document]
QuintonC: > Can you clarify this a bit? Lots of the word "match" and I am not sure exactly what it is saying.

Yep! This was copied over, but I can fine-tune this comment a bit!

packages/connect-wallet/src/components/Modal/Modal.tsx
```diff
@@ -36,36 +39,57 @@ import {
...
+  const {disconnect} = useDisconnect();
   const isMounted = useIsMounted();
   const escPress = useKeyPress('Escape');
   const [ref, {height}] = useMeasure();
   const isSmall = useMediaQuery('smDown');
-  const {active, closeModal, navigation} = useModal();
   const reducedMotion = useReducedMotion();
   const {t} = useTranslation('Modal');
 
+  useMiddleware({enableDelegateCash, orderAttributionMode, requireSignature});
```
caropinzonsilva, line 61: Just curious, where was this done before?

packages/connect-wallet/src/components/Modal/Screens/SignatureScreen.tsx
```diff
@@ -2,25 +2,28 @@ import {eventNames} from '@shopify/blockchain-components';
...
+  const {error, signing} = useAppSelector((state) => state.modal);
   const {pendingWallet} = useAppSelector((state) => state.wallet);
-  const {clearError, closeModal, error, signing, requestSignature} = useModal();
+  const {signMessage} = useSignMessage();
   const {t} = useTranslation('Screens');
 
   const handleSignMessage = useCallback(() => {
     if (!pendingWallet) {
-      closeModal();
+      dispatch(closeModal());
```
caropinzonsilva, line 20: Since we are in the signature screen, do we need to also disconnect like we do in the `Modal`? https://github.com/Shopify/blockchain-components/pull/142/files#diff-634b68577243b1fb1ec391d706267033365225d541028f4680b459572c7d5e9cR72

caropinzonsilva (approved):
Looks good. Approving so that it is not blocking the merge, but I left 2 questions

packages/connect-wallet/src/components/Modal/Modal.tsx
QuintonC: > Just curious, where was this done before?

This was done in the `ModalProvider` ([seen here](https://github.com/Shopify/blockchain-components/pull/142/files#diff-3144bd5e85f5586dd3d8e655b3b67cbd4552ca40c117527f4f38b048839f40f5L215))

packages/connect-wallet/src/components/Modal/Screens/SignatureScreen.tsx
QuintonC: Good question... I'm not sure that this should / would be possible but I think the previous implementation covered this edge case. I'm going to test this out real quick and see if we need to add the disconnect logic or not.

QuintonC: I took another look at it and I'm going to add the disconnect call here!

I'm not sure that it will necessarily do anything other than self-correct an odd client state, which is enough of a reason to add it back to me. 

Good catch!

PR #145, opened 2023-03-20
[delegate-cash] add icon in ConnectButton

caropinzonsilva:

In this PR, we add the delegate cash icon with the number of vaults inside the `ConnectButton`

## 🕹️ Demonstration

20-27-jy3o2-hz02x

- Connect hot wallet
- If wallet has delegations, the delegate cash icon should appear in the Connect Button
- If the wallet has delegations, the number of vaults should appear next to the delegate cash icon

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const react = require('@vitejs/plugin-react');
+const {nodePolyfills} = require('vite-plugin-node-polyfills');
 
 module.exports = {
   stories: [
@@ -15,7 +16,10 @@ module.exports = {
     options: {},
   },
   async viteFinal(config) {
-    config.plugins = mergePlugins(config.plugins, react());
+    config.plugins = mergePlugins(config.plugins, [
+      nodePolyfills(),
+      ...react(),
+    ]);
     return config;
   },
 };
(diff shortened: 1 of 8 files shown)


jamiely (approved):

packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -105,6 +106,14 @@ export const ConnectButton = () => {
         <Text as="span" variant="bodyLg">
           {buttonLabel}
         </Text>
+        {vaults && vaults.length > 0 ? (
+          <div className="sbc-flex sbc-flex-row sbc-items-center sbc-justify-center sbc-gap-x-1 sbc-px-2">
+            <div className="sbc-flex sbc-h-3">{DelegateCash}</div>
```
QuintonC, line 111: I think we can omit the `sbc-flex` class on this element since it does not need to have `display: flex;`

QuintonC (commented):
It looks like this merged before I got a chance to review it, but I left some feedback anyway.

PR #147, opened 2023-03-21
[delegate-cash] Add VaultList

caropinzonsilva:

Adds the list of vaults inside the `ConnectButton` popover.

## 🕹️ Demonstration
21-35-7lg92-j6isq

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/connect-wallet/src/components/VaultsList/VaultsList.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import {WalletAddressRow} from './components/VaultsListRow';
+
+export const VaultsList = ({vaults}: {vaults?: Address[]}) => {
+  const {t} = useTranslation('VaultsList');
+
+  if (!vaults) return null;
+
+  return (
+    <div className="sbc-w-full sbc-rounded sbc-bg-address-chip sbc-p-4">
```
caropinzonsilva, line 14: I decided to use the class `sbc-bg-address-chip` instead of adding a new CSS variable specifically for the `VaultList`. My reasoning behind this is that it makes sense for them to have similar styling: they are both in the same popover, and they both have the copy+paste wallet interaction.
21-47-6mju5-2liil

packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -142,6 +143,8 @@ export const ConnectButton = () => {
           {copied ? CircleTick : Copy}
         </button>
 
+        <VaultsList vaults={vaults} />
```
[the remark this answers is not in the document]
QuintonC: It looks like `VaultsList` takes care of the render and renders null when vaults does not have any length 👍

packages/connect-wallet/src/components/VaultsList/VaultsList.tsx
QuintonC: Can we change the `sbc-rounded` to `sbc-rounded-lg` to match the designs?

You might have to add the class to the `borderRadius` definitions in `tailwind-config/tailwind.config.js`

packages/connect-wallet/src/components/VaultsList/components/VaultsListRow.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const WalletAddressRow = ({vault}: Props) => {
+  const {copy, copied} = useCopyToClipboard();
+
+  return (
+    <button
+      onClick={() => copy(vault)}
+      type="button"
+      className="sbc-flex sbc-w-full sbc-cursor-pointer sbc-items-center sbc-justify-between sbc-gap-x-3 sbc-bg-address-chip sbc-py-2 sbc-px-0 sbc-text-address-chip sbc-transition-colors sbc-border-none hover:sbc-bg-address-chip-hover"
```
QuintonC, line 16: I'm unsure about the `background-color` here. Something about the duplicated background colors feels a bit strange to me. Can we clarify this with design? I'm wondering if an opacity transition on hover might be more fitting here (e.g. `sbc-transition-opacity hover:sbc-opacity-50`)

If we want to retain background-color we might want to check if we should add padding to the rows so that the background color change isn't edge to edge with the text sitting against the edge.

packages/connect-wallet/src/providers/I18nProvider/translations/en/VaultsList.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title_one": "Delegate for {{count}} wallet",
+  "title_other": "Delegate for {{count}} wallets"
+}
```
QuintonC, line 4: 👏  nice!

QuintonC (approved):
Left a couple of comments. 

One of them we should probably clarify with design on.

packages/connect-wallet/src/components/VaultsList/VaultsList.tsx
lesliexin: That's a good point!

packages/connect-wallet/src/providers/I18nProvider/translations/en/VaultsList.json
lesliexin, line 4: Ooh nice to know that `i18nNext` also supports pluralization!

packages/connect-wallet/src/components/VaultsList/components/VaultsListRow.tsx
caropinzonsilva: > Something about the duplicated background colors feels a bit strange to me. 
Can I ask you why? We have duplicated background colors in all our button components. for example, this is the `copy wallet address` button:
21-24-e6gt6-ommt8

> we should add padding to the rows so that the background color change isn't edge to edge with the text sitting against the edge.
Sure! I can ask design about this.

packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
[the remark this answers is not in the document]
caropinzonsilva: Sure! I can add the `enableDelegateCash` logic here, but do you think that we would also want to add the `vault.length > 0` condition here too? In my opinion, the parent should only know that it needs to render the list, the list itself should know that it should not render if there are no items 🤔

packages/connect-wallet/src/components/VaultsList/VaultsList.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <div className="sbc-flex sbc-items-start sbc-gap-x-4">
+        <div className="sbc-w-4">{DelegateCash}</div>
+        <Text variant="bodyMd">{t('title', {count: vaults.length})}</Text>
+      </div>
+      <div className="sbc-flex sbc-flex-col sbc-pt-3">
+        {vaults.map((vault, index) => (
+          <div key={vault} className="sbc-w-full">
+            <WalletAddressRow vault={vault} />
+            {index < vaults.length - 1 ? (
```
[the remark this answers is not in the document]
caropinzonsilva: Oh, interesting! I did not, but I'm going to refactor it so that it uses CSS. I prefer not adding new elements when possible ✨

packages/connect-wallet/src/components/VaultsList/components/VaultsListRow.tsx
QuintonC: > Can I ask you why? We have duplicated background colors in all our button components. for example, this is the copy wallet address button

Hmm... maybe "duplicated background colors" wasn't the correct phrasing here. 

What I meant was the `VaultList` component has the same background color as the `VaultListRow`. If we want the VaultListRow to have a background color on hover I'd suggest maybe making it look more like a button by giving it some padding.

caropinzonsilva: Sure! I can add that padding. I think it is important for the element to have a hover effect for improving accessibility.

packages/connect-wallet/src/components/VaultsList/VaultsList.tsx
```diff
@@ -0,0 +1,31 @@
+import {Text, DelegateCash} from 'shared';
+import {Address} from 'wagmi';
+
+import {useTranslation} from '../../hooks/useTranslation';
+
+import {WalletAddressRow} from './components/VaultsListRow';
+
+export const VaultsList = ({vaults}: {vaults?: Address[]}) => {
```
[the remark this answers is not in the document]
caropinzonsilva: Addressed here: https://github.com/Shopify/blockchain-components/pull/147/commits/3ec0db408b9a6faa213e2a15a8df2f4092756708

packages/tailwind-config/tailwind.config.js
```diff
@@ -164,6 +164,15 @@ module.exports = {
...
         },
+        '.border-t-0': {
+          'border-top-style': 'unset',
+        },
+        '.border-l-0': {
+          'border-left-style': 'unset',
+        },
+        '.border-r-0': {
+          'border-right-style': 'unset',
+        },
```
caropinzonsilva, line 175: Since we have preflight disabled, I had to manually add these classes. I'm following the Tailwind naming conventions: https://tailwindcss.com/docs/border-width

QuintonC: This is a bit confusing to me because the `border-t-0` would be for `border-width` but you're adding them to `borderStyle`.

Did you mean to add styles for `border-width` or `border-style`?

QuintonC: Pulled this code into my local environment and took another look and it made more sense. However, I'd recommend not mixing syntax between border-style and border-width.

I'd recommend either moving `theme.borderWidth` into `theme.extend.borderWidth` within `tailwind.config.js` or adding the defaultTheme.borderWidth[0] to `theme.borderWidth`

```diff
    borderWidth: {
      'button-disabled': 'var(--sbc-border-width-button-disabled, 0px)',
      'button-primary': 'var(--sbc-border-width-button-primary, 0px)',
      'button-secondary': 'var(--sbc-border-width-button-secondary, 1px)',
      divider: 'var(--sbc-border-width-divider, 1px)',
      popover: 'var(--sbc-border-width-popover, 0px)',
      tokengate: 'var(--sbc-border-width-tokengate, 1px)',
+     0: defaultTheme.borderWidth[0],
    },
```

Either of these approaches will allow you to shorten the `className` in `VaultRow.tsx` by replacing `sbc-border-t-0 sbc-border-l-0 sbc-border-r-0` with `sbc-border-0`

packages/connect-wallet/src/components/ConnectButton/ConnectButton.tsx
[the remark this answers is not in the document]
caropinzonsilva: No problem at all! Thanks for commenting in the PR and helping me make it better:)

packages/tailwind-config/tailwind.config.js
caropinzonsilva: Addressed here! https://github.com/Shopify/blockchain-components/pull/147/commits/5304e76c7658f5fb7e10f78aaad504ac280fca85 It is so much better, thank you

PR #149, opened 2023-03-21
[delegate-cash] Send vaults on onConnect callback

caropinzonsilva:

The `vaults` were already sent before, so this PR just makes this visible in the playground example.

## 🕹️ Demonstration
image

- Run locally the playground `yarn dev`
- Connect a wallet with delegations
- See console log of `checkIfWalletMeetsRequirements` event that includes the vaults

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/apps/playground/src/types/events.ts
+++ b/apps/playground/src/types/events.ts
@@ -34,6 +34,7 @@ export interface CheckIfWalletMeetsRequirementsEvent extends EventBusEvent {
     address: string;
     message: string;
     signature: string;
+    vaults?: string[];
   };
   response: {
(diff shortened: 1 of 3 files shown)


apps/playground/src/types/events.ts
```diff
@@ -34,6 +34,7 @@ export interface CheckIfWalletMeetsRequirementsEvent extends EventBusEvent {
     address: string;
     message: string;
     signature: string;
+    vaults: string[];
```
lesliexin, line 37: Should this be optional?

jamiely (approved):

PR #158, opened 2023-03-27
[delegate-cash] Refactor connect wallet logic to send vaults on connect

caropinzonsilva:
We noticed that the `vaults` were undefined 

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] Tested on mobile
- [ ] Includes unit tests


packages/connect-wallet/src/hooks/useConnectWallet/useConnectWalletCallbacks.ts
```diff
@@ -12,15 +11,24 @@ export const useConnectWalletCallbacks = (props?: useConnectWalletProps) => {
...
   useEffect(() => {
-    const listener = buildOnConnectMiddleware(({wallet}) => {
-      publishEvent(eventNames.CONNECT_WALLET_ON_CONNECT_EVENT, {
-        address: wallet.address,
-        vaults: wallet.vaults,
-        connector: wallet.connectorId,
-      });
-      onConnect?.(wallet);
+    const listener = addListener({
+      actionCreator: attributeOrder.fulfilled,
```
caropinzonsilva, line 22: We are now considering the connect wallet flow done after the order is attributed. We could add an extra step or rename the `attributeOrder` action in the future so that the connection between attributing an order and finishing the connect wallet flow is more clear, but I decided to not touch this in this PR.

packages/connect-wallet/src/hooks/useMiddleware.ts
```diff
@@ -164,23 +166,13 @@ export const useMiddleware = ({
...
-        state.dispatch(
-          attributeOrder({
-            orderAttributionMode,
-            wallet,
-          }),
-        );
-      }
+      state.dispatch(
+        fetchDelegations({walletAddress: wallet.address, enableDelegateCash}),
+      );
```
caropinzonsilva, line 175: Instead of having this conditional action flow here, we modified it to always call the same actions.

packages/connect-wallet/src/slices/walletSlice/delegateCash.ts
```diff
@@ -4,7 +4,20 @@ import {Address} from 'wagmi';
...
   'wallet/fetchDelegations',
-  async (walletAddress: string, thunkApi) => {
+  async (
+    {
+      walletAddress,
+      enableDelegateCash,
+    }: {walletAddress: Address; enableDelegateCash?: boolean},
+    thunkApi,
+  ) => {
+    if (!enableDelegateCash) {
```
caropinzonsilva, line 14: Instead of conditionally calling this action, we always call it and return `undefined` for the vaults if `enableDelegateCash` is false.

packages/connect-wallet/src/hooks/useMiddleware.ts
```diff
@@ -126,6 +126,8 @@ export const useMiddleware = ({
    *
    * This listener will run after the delegations are fetched and
    * will attribute the wallet addresses to the order.
+   *
+   * Complete flow diagram of connecting a wallet: https://tinyurl.com/2p8dh33d
```
jamiely, line 130: This is good as-is. Whipped this up in case you want to embed it:

```
                ┌─┐
                │O│
                └┬┘
                 │
     ┌───────────▼──────────────────┐
     │           │                  │
┌────▼────┐ ┌────▼────┐ ┌───────────▼──────────┐
│addWallet│ │setActive│ │validatePendingWallets│
└────┬────┘ └────┬────┘ └───────────┬──────────┘
     │           │                  │
     └───────────┴──┬───────────────┘
                    │
           ┌────────▼──────────┐
           │onConnectMiddleware├───────┐
           └────────┬──────────┘       │
                    │                  │
             ┌──────▼─────────┐        │
             │fetchDelegations│        │
             └──────┬─────────┘        │
                    │                  │
             ┌──────▼─────────┐        │
             │orderAttribution│        │
             └──────┬─────────┘        │
                    │                  │
          ┌─────────▼────────────┐ ┌───▼────┐
          │connectWalletCallbacks│ │fetchEns│
          └─────────┬────────────┘ └───┬────┘
                    │                  │
                   ┌▼┐                 │
                   │O│◄────────────────┘
                   └─┘
```

PR #159, opened 2023-03-28
[tailwind-config] Use pixels instead of REMs

caropinzonsilva:
Update Tailwind config to use pixels instead of REMs. We noticed that if the components were included in an application that has smaller or larger root font-size, the styling broke. We made the decision to use pixels instead.

## 🕹️ Demonstration
Before:
28-41-y82da-c6c35

Now:
28-48-gqgpa-t374e

Can be 🎩 in https://commercetown.xyz/products/townfolk-103-print

## ✅ Checklist

- [x] Tested on mobile


lesliexin (approved):
Thank you!

PR #161, opened 2023-03-29
Add V1.0.0 for dependency packages

caropinzonsilva:

Changeset for releasing V1.0.0 for @shopify/blockchain-components and @shopify/gate-context-client


lesliexin (approved):
Thank you!

PR #163, opened 2023-04-06
:sparkles: Adds support for active.end / end date value

QuintonC:

Adds support for `active.end` within the `Tokengate` component.

During development of this I realized there was some similarity between the component which would display at the bottom of the card via the following states: 

- `TokengateCardSection.ConnectWallet`
- `TokengateCardSection.AvailableSoon`
- `TokengateCardSection.SoldOut`

Since these are all so similar, I found that it would make sense to create a wrapper component that accepted some similar props for translation key, value, and namespace. Though they all share the same namespace now, I left the namespace open so that it could be more dynamic in the future.

This also means we can create a single component to maintain the spacing and colors between all of these different sections.

Another thing that popped up was that the translation utilities from react-i18next make the translation string with dates include `at` instead of a comma between the date and the time.

## 🕹️ Demonstration

| Name / State | 📖 Storybook Link | Image |
| - | - | - |
| Available Soon | | |
| End Date | | |
| Start Date | | |
| Sold Out | | |

I believe the following two sections could probably be converted to use the new wrapper as well, but I omitted doing so since it would involve adjusting the logic of the `Tokengate` component which I do not want to do in this pull request.

- `TokengateCardSection.OrderLimitReachedError`
- `TokengateCardSection.MissingTokensError`

Use the storybook links mentioned in the table above.

## ✅ Checklist

- [ ] ~Tested on mobile~
- [ ] ~Tested on multiple browsers~
- [ ] ~Tested for accessibility~
- [x] Includes unit tests
    - Will require an update to dev docs to change type of button from `ReactNode` to `JSX.Element`


packages/tokengate/src/components/Tokengate/utils.ts
```diff
@@ -1,6 +1,11 @@
 import {useTranslation} from '../../hooks/useTranslation';
 import {TokengateProps, UnlockingToken} from '../../types';
 
+export type UtilsProps = Omit<
+  TokengateProps,
+  'connectButton' | 'connectedButton'
+>;
```
QuintonC, line 7: Since `connectButton` and `connectedButton` were converted to `JSX.Element` type we would need to convert the test file to `utils.spec.tsx` to support the rendering of a `JSX.Element` in the test file.

The reason that connectButton and connectedButton were changed was to have a stricter type check in `ButtonWrapper` when checking if the provided `button` prop was a `string` or `object`.

PR #169, opened 2023-04-13
[connect-wallet] Add modal header custom title

caropinzonsilva:
⚠️ Fixes: Part of https://github.com/Shopify/nft-gating/issues/1374

Add modal header custom title.

## 🕹️ Demonstration
13-08-2sc8o-31fkd

Can be 🎩  in https://639b1f308693132693d9b82c-zgloulwnjs.chromatic.com/?path=/story/connect-wallet--playground&args=modalHeader:Connect+Wallet+test

- In the story controls, change the `modalHeader`
- Open the connect-wallet modal by clicking the story button
- The new header should be reflected in the modal

## ✅ Checklist

- [x] Tested on mobile
- [x] Includes unit tests
- [x] Updated relevant documentation for the changes (if necessary)


packages/connect-wallet/src/providers/ConnectWalletProvider/context.tsx
```diff
@@ -9,6 +9,9 @@ import {StatementGenerator} from './types';
 export interface ConnectWalletProviderValue {
   chains: Chain[];
   connectors: Connector[];
+  customTitles?: {
+    modalHeader?: string;
```
caropinzonsilva, line 13: This naming was challenging. If we want to be extra explicit, it should be `modalConnectScreenHeader` since we only show it in the `ConnectScreen`. However, so far we do not want to customize any other headers, so I thought maybe this naming will be enough. But I'm happy to change it to a better option.

jamiely (approved):
:+1:

PR #174, opened 2023-04-26
Format discount values

caropinzonsilva:

Format the discount values and adds support for strings. Like this, no matter if the discount value is wrongly formatted, we display the correct format in the Tokengate component.

## 🕹️ Demonstration

26-24-fhlgi-4dm43

Can be 🎩 in https://639b1f308693132693d9b82c-qzepolaquu.chromatic.com/?path=/story/tokengate-discount-locked--locked

1. The value of the discount should be `$10.00`
1. In the controls, change the `reaction.discount.type` to `percentage`
1. The value of the discount should be `10%`
1. In the controls, change the `reaction.discount.value` to `"10.0"
1. The value of the discount should be `10%`
1. In the controls, change the `reaction.discount.type` to `fixedAmount`
1. The value of the discount should be `$10.00`

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests

### DIFF
--- a/.changeset/spicy-dolls-pay.md
+++ b/.changeset/spicy-dolls-pay.md
@@ -1,0 +1,7 @@
+---
+'@shopify/tokengate': patch
+---
+
+Adds formatting to the discount value. If the discount is type `fixedAmount`, we format to two decimal places. If the discount is type `percentage`, we remove any decimals.
+
+Adds support for `string` values to the reaction type discounts.
--- a/packages/tokengate/src/components/Tokengate/utils.spec.ts
+++ b/packages/tokengate/src/components/Tokengate/utils.spec.ts
@@ -121,7 +121,45 @@ describe('Tokengate - utils', () => {
             useTitleAndSubtitle({...defaultDiscountTokengateProps}),
           );
 
-          expect(result.current.title).toBe('$10 holder discount');
+          expect(result.current.title).toBe('$10.00 holder discount');
+          expect(result.current.subtitle).toBe(
+            'To unlock this discount, you need:',
+          );
+        });
+
+        it('correctly formats the value for fixedAmount discount type', () => {
+          const {result} = renderHook(() =>
+            useTitleAndSubtitle({
+              ...defaultDiscountTokengateProps,
+              reaction: DiscountReactionFixture({
+                discount: {
+                  type: 'fixedAmount',
+                  value: '10.0',
+                },
+              }),
+            }),
+          );
+
+          expect(result.current.title).toBe('$10.00 holder discount');
+          expect(result.current.subtitle).toBe(
+            'To unlock this discount, you need:',
+          );
+        });
+
+        it('correctly formats the value for percentage discount type', () => {
+          const {result} = renderHook(() =>
+            useTitleAndSubtitle({
+              ...defaultDiscountTokengateProps,
+              reaction: DiscountReactionFixture({
+                discount: {
+                  type: 'percentage',
+                  value: '10.0',
+                },
+              }),
+            }),
+          );
+
+          expect(result.current.title).toBe('10% holder discount');
           expect(result.current.subtitle).toBe(
             'To unlock this discount, you need:',
           );
@@ -149,7 +187,7 @@ describe('Tokengate - utils', () => {
               isLocked: false,
             }),
           );
-          expect(result.current.title).toBe('$10 discount unlocked');
+          expect(result.current.title).toBe('$10.00 discount unlocked');
           expect(result.current.subtitle).toBe(
             'Your token got you access to this discount!',
           );
@@ -187,7 +225,7 @@ describe('Tokengate - utils', () => {
               ],
             }),
           );
-          expect(result.current.title).toBe('$10 discount unlocked');
+          expect(result.current.title).toBe('$10.00 discount unlocked');
           expect(result.current.subtitle).toBe(
             'You can use this discount up to 4 times with your tokens.',
           );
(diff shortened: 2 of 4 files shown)


packages/blockchain-components/tsup.config.bundled_p0q2alxmg9o.mjs
```diff
@@ -0,0 +1,21 @@
+// tsup.config.ts
```
QuintonC, line 1: Can this file be removed? 🤔

packages/tokengate/src/components/Tokengate/utils.spec.ts
```diff
@@ -121,7 +121,45 @@ describe('Tokengate - utils', () => {
...
           );
 
-          expect(result.current.title).toBe('$10 holder discount');
+          expect(result.current.title).toBe('$10.00 holder discount');
+          expect(result.current.subtitle).toBe(
+            'To unlock this discount, you need:',
+          );
+        });
+
+        it('formats correctly the fixedAmount discount values', () => {
```
QuintonC, line 130:
```suggestion
        it('correctly formats the value for fixedAmount discount type', () => {
```

packages/tokengate/src/components/Tokengate/utils.spec.ts
```diff
@@ -121,7 +121,45 @@ describe('Tokengate - utils', () => {
...
+            }),
+          );
+
+          expect(result.current.title).toBe('$10.00 holder discount');
+          expect(result.current.subtitle).toBe(
+            'To unlock this discount, you need:',
+          );
+        });
+
+        it('formats correctly the percentage discount values', () => {
```
QuintonC, line 149:
```suggestion
        it('correctly formats the value for percentage discount type', () => {
```

packages/tokengate/src/types.ts
```diff
@@ -66,7 +66,7 @@ interface DiscountReaction {
   type: 'discount';
   discount: {
     type: 'fixedAmount' | 'percentage';
-    value: number;
+    value: number | string;
```
QuintonC, line 69: Just a reminder that this will need to be updated on .dev docs

packages/blockchain-components/tsup.config.bundled_p0q2alxmg9o.mjs
jamiely, line 1: this shouldn't be committed, should it?

jamiely (approved):

packages/tokengate/src/components/Tokengate/utils.ts
```diff
@@ -1,6 +1,11 @@
 import {useTranslation} from '../../hooks/useTranslation';
 import {TokengateProps, UnlockingToken} from '../../types';
 
+const formatter = new Intl.NumberFormat('en-US', {
+  style: 'currency',
+  currency: 'USD',
+});
```
QuintonC, line 7: No change requested here, just thinking out loud.

I think it will be interesting to see how we load translations / formatting for the numbers here if we being to explore the functionality of adding / loading translations into this package. 

I was looking into some of the work that Polaris is working on with adding / loading translations with i18next and thinking about how we will probably want to bring some similar functionality to our components where folks can provide their own translation dictionaries for the components.

packages/tokengate/src/components/Tokengate/utils.spec.ts
```diff
@@ -187,7 +225,7 @@ describe('Tokengate - utils', () => {
               ],
             }),
           );
-          expect(result.current.title).toBe('$10 discount unlocked');
+          expect(result.current.title).toBe('$10.00 discount unlocked');
```
jamiely, line 228: This looked nicer without cents

packages/blockchain-components/tsup.config.bundled_p0q2alxmg9o.mjs
caropinzonsilva: It should not! Thanks for noticing

packages/blockchain-components/tsup.config.bundled_p0q2alxmg9o.mjs
caropinzonsilva: Yep!

packages/tokengate/src/components/Tokengate/utils.ts
caropinzonsilva: I agree! We can look at that in the future when we add more currencies, although I really liked the `Intl.NumberFormat` function: no libraries needed and it was really straightforward to use!

packages/tokengate/src/types.ts
caropinzonsilva: Sure! I just the `string` it since this is what we were sending in the gates from the BE

.changeset/spicy-dolls-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/tokengate': patch
+---
+
+Fix discount values display
```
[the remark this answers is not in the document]
caropinzonsilva: Sure!

packages/tokengate/src/components/Tokengate/utils.spec.ts
caropinzonsilva: I know, but I guess having two cents is a Shopify standard :/

packages/tokengate/src/types.ts
QuintonC: Yep, I think we can push that change to the dot.dev content prior to this ships if you'd like to as well.

packages/tokengate/src/components/Tokengate/utils.ts
QuintonC: 100% 😄 

Definitely cool to see how this all comes together with a package and translations being provided.

.changeset/spicy-dolls-pay.md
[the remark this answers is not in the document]
QuintonC: Thank you! 🙌

QuintonC (approved):
Great work! Thanks for addressing comments 👏 

🚢

PR #185, opened 2023-05-11
:sparkles: Adds support for custom titles with dynamic values

QuintonC:
⚠️ Fixes: 

This PR adds support for custom `discountTitles` that retain the discount value within the title. This will work for both `locked` and `unlocked` title variants. 

This also relocates some code that was used for tokengate state / titles, etc. The files were quite large and becoming cumbersome to comb through while attempting to add support for new features.

## 🕹️ Demonstration

...

...

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A
- [x] Includes unit tests


c-bacon (approved):
🔥

jamiely (approved):
LGTM

This is good as-is. Some of these changes seem like a matter of preference rather than needing to be done to test/clean up the code. I only mention this b/c you mentioned this PR being a bigger one.

packages/tokengate/src/hooks/useTokengateHeadings.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  if (customTitle && typeof customTitle === 'string') {
+    i18n.addResource('en', 'Tokengate', 'customTitle', customTitle);
+  }
+
+  const title = customTitle
+    ? t('customTitle', {value: discount})
+    : t(`${i18nKeyPrefix}.title`, {value: discount});
+
+  let subtitle = customSubtitle || t(`${i18nKeyPrefix}.subtitle`);
```
lesliexin, line 69: Just out of curiosity, why don't we support the use of {{value}} in the subtitle as well? To me, if it's possible in the title, I would expect that it's also possible in the subtitle.

lesliexin (approved):
Thanks for updating storybook as well, made it super easy to tophat!

packages/tokengate/src/hooks/useTokengateHeadings.ts
QuintonC: > Just out of curiosity, why don't we support the use of {{value}} in the subtitle as well? To me, if it's possible in the title, I would expect that it's also possible in the subtitle.

Hmm... good question. I don't know that there are any dynamic values that get inserted into the titles. We have the ability to provide custom titles when there is an `orderLimit`, but I don't know that we've run into a case that uses it just yet. 

Maybe we can keep an eye on that and add support for it in the future 🤔 

Good question!

lesliexin: Makes sense, thanks!!

PR #186, opened 2023-05-11
feat: add imToken Connector

izayl:

add [imToken](https://token.im) connector

## 🕹️ Demonstration

image

...

...

## ✅ Checklist

- [x] Tested on mobile
- [ ] Includes unit tests


izayl:
Thank you very much for your reply. Personally, I think providing options for deep integration with major wallets would be very helpful for developers and users, as [imToken](https://token.im) has already served over 12 million users. However, we fully understand that this also depends on your team, and we respect your team's decision.

QuintonC:
> I think providing options for deep integration with major wallets would be very helpful for developers and users

Definitely! Initially we **only** supported the list of default connectors, but later added functionality for folks to provide custom connectors to their configuration.

If you think there are any features we can provide that make creating custom connectors easier, feel free to let us know 🙂

PR #190, opened 2023-05-16
🔇 Removes debug logs for i18n context in non-dev runtimes

QuintonC:

Utilizes `process.env.NODE_ENV` to conditionally enable the `debug` flag for i18n context. This will remove the logging that we previously saw in the console relating to i18n context.

## 🕹️ Demonstration

| Before | After |
| - | - |
| Screenshot 2023-05-16 at 4 25 59 PM | Screenshot 2023-05-16 at 4 42 09 PM |

1. Pull main.
2. `yarn install`
3. `yarn example`
  a. Run any example
4. When the app launches you will see logging from the connect-wallet package.

Repeat the steps above, but instead of pulling main, pull this branch. 

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/packages/tokengate/src/providers/I18nProvider/i18n.ts
+++ b/packages/tokengate/src/providers/I18nProvider/i18n.ts
@@ -11,7 +11,9 @@ import TokengateRequirements from './translations/en/TokengateRequirements.json'
 
 const i18n = createInstance({
   fallbackLng: 'en',
-  debug: true,
+  // Only enable the debug flag in development environments.
+  // eslint-disable-next-line no-process-env
+  debug: process.env.NODE_ENV === 'development',
 
   resources: {
     en: {
--- a/.changeset/seven-gorillas-sparkle.md
+++ b/.changeset/seven-gorillas-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@shopify/connect-wallet': patch
+'@shopify/tokengate': patch
+---
+
+Removed i18n context debug logging in non-development environments.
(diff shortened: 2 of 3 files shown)


packages/tokengate/src/providers/I18nProvider/i18n.ts
```diff
@@ -11,7 +11,9 @@ import TokengateRequirements from './translations/en/TokengateRequirements.json'
 
 const i18n = createInstance({
   fallbackLng: 'en',
-  debug: true,
+  // Only enable the debug flag in development environments.
+  // eslint-disable-next-line no-process-env
```
Soleone, line 15: So if I understand correctly this eslint rule is there because there is a different approach on how to realize this different behavior in different environments?

I know in Rails we don't want to do this. Instead there is a dedicated production.rb vs development environment file where you can set "global" variables like "debug = true" vs false and then depending on the environment the right file is loaded and all variables set up.

Is there not something similar available here?

QuintonC: Good question.

This is more of a stylistic issue for ESLint ([you can view the rule here](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-process-env.md))

There really isn't a good alternative IMO because the alternative would be to load a `config` file which has a conditional in it that still has to use a variable from `process.env` for the current environment. IMO, the rule can probably be disabled, or at least, there could be an allowlist for the rule to work in _certain_ files.

It really only needs to be used in the `i18n.ts` file inside of `packages/tokengate` and `packages/connect-wallet` and the `configureStore.ts` file inside of `packages/connect-wallet`.

Soleone (approved):

PR #193, opened 2023-05-17
:bug: Address issue with preflight button style collision

QuintonC:

Addresses an issue where Tailwind preflight styles were colliding with button styles.

## 🕹️ Demonstration

| Before | After |
| - | - |
| SCR-20230517-n43 | SCR-20230517-n2t |

It's a bit complex, but this is being tested in a headless environment with Hydrogen which requires another patch in order to function.

You can try installing in another project that is using Tailwind and include the preflight styles to ensure that the styles for the button work as expected.

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A

### DIFF
--- a/packages/connect-wallet/src/styles.css
+++ b/packages/connect-wallet/src/styles.css
@@ -1,3 +1,4 @@
+@import '../../shared/src/components/Button/Button.css';
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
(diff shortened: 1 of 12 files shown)


packages/connect-wallet/src/styles.css
```diff
@@ -1,3 +1,4 @@
+@import '../../shared/src/components/Button/Button.css';
```
QuintonC, line 1: Imports must occur prior to any `@tailwind` directives. Additionally, we have to use relative path importing here it seems.

This does _not_ show up in the compiled CSS as a relative import (or an import at all).

Screenshot 2023-05-17 at 4 27 01 PM

jamiely (approved):
:+1:

Issue #194, opened 2023-05-17
Remove usage of redux-persist


QuintonC:
While initially attempting to utilize `@shopify/connect-wallet` in a hydrogen storefront, I ran into the following error

```log
Error: Some functionality, such as asynchronous I/O (fetch, Cache API, KV), timeouts (setTimeout, setInterval), and generating random values (crypto.getRandomValues, crypto.subtle.generateKey), can only be performed while handling a request.
```

This error was later tracked down to our usage of `redux-persist` in the `@shopify/connect-wallet` package. Because the `persistReducer` method is using `setTimeout` during rehydration, we receive this error from miniflare via Oxygen.

This will need to be fixed in a more clever way so that we can continue to rely on persistent data being present in our package. Persistence is important for this package to ensure that connection state is maintained across sessions (whether those are shopping sessions, browser sessions, or even page refreshes).

Issue #46, opened 2023-03-03
Track Redux Toolkit + Redux ESM support releases


QuintonC:
## Description

During the process of resolving issues with use of the package in webpack based bundlers we made an upgrade to utilize v2 of RTK which also requires a bump to v5+ of Redux.

Both are in alpha and this issue is primarily to ensure that we continue to track the release of RTK v2 and Redux v5. The expected outcome of this PR would be that we continue to track and implement changes from the v2 and v5 releases to ensure that our package can be used in additional frameworks + bundler combinations.

## Related
Redux PR: https://github.com/reduxjs/redux/pull/4491
Redux Toolkit PR: https://github.com/reduxjs/redux-toolkit/pull/3089
RTK + Redux Alpha upgrade PR: https://github.com/Shopify/blockchain-components/pull/37

PR #196, opened 2023-05-19
Chore: Remove Redux

QuintonC:
⚠️ Closes: https://github.com/Shopify/blockchain-components/issues/46
⚠️ Closes: https://github.com/Shopify/blockchain-components/issues/194

While doing testing in headless storefronts I came across a unique issue that has brought up a desire to test out something.

The end goal is to be able to place the `ConnectWalletProvider` in the `root.tsx` of a Hydrogen storefront, make use of `useConnectWallet` inside of the `Layout.tsx` file and when a wallet is connected (i.e. use `onConnect` callbacks), perform a `POST` request to store the wallet data inside of the storefront session. 

This will ensure that the session stays in sync with our storage of the connected wallet and also allow us to access wallet data across the entire storefront, resulting in the possibility to gate any subject on a headless storefront.

With Redux, doing this results in an error of the `useConnectWallet` hook being used outside of the context of a Redux provider. We need to be able to read this data and persist it on the server side (if it is available) in order to create an efficient gating implementation with Hydrogen.

Some side effects of this (aside from a rather large amount of code being changed), is that the package _should_ be smaller since Zustand is much smaller than Redux.

Additionally, something of note is that the callback provided to useConnectWallet for onConnect will run earlier than it previously did. It previously waited for fetchEns and fetchDelegations to resolve before running the callback. Now, the provided function runs immediately when an `activeWallet` has been set from `undefined` to a defined `Wallet` type.

This can be tophatted by using the examples in the repository or by using the playground. Either of the two are suitable for testing the changes in this pull request.

## ✅ Checklist

- [x] Tested on mobile
- [x] Includes unit tests


QuintonC:
Rebased on main to include the most recent changes (`CODEOWNERS` update) and to satisfy a merge requirement.

QuintonC:
> Amazing work! I tophatted both in mobile and in desktop, deleting the cookies and localstorage, and it all worked as expected 🚀

Thank you so much! 🙌

PR #197, opened 2023-05-22
Removes some folks from codeowners

jamiely:
Removes folks from codeowners


QuintonC (approved):
Just to clarify, @c-bacon wants to stay on the reviews for this?

PR #203, opened 2023-06-15
:bug: Address issue with lack of keys

QuintonC:

While working with the Tokengate component in a Hydrogen setting, I noticed that there was a violation of lack of unique keys in UnlockingTokens. When I went to resolve the issue I noticed that we were mapping through collections two times before rendering in a `TokenList`. This changes that implementation to map a single time and return an array of elements instead of an array of objects which are then mapped over to return an array of elements.

## 🕹️ Demonstration

Errors from before:
SCR-20230615-ol4

I'm no longer getting the error when testing the component in Hydrogen.

## ✅ Checklist

- [ ] ~Tested on mobile~ N/A


packages/tokengate/src/components/TokenList/TokenListSkeleton.tsx
```diff
@@ -1,13 +1,17 @@
...
+export const TokenListSkeleton = () => (
+  <div className="sbc-relative">
+    <div className="sbc-flex sbc-w-full sbc-flex-row sbc-items-center sbc-gap-x-3">
+      <div className="sbc-relative sbc-h-12 sbc-w-12">
+        <div className="sbc-h-full sbc-w-full sbc-overflow-hidden sbc-rounded-full">
+          <SkeletonThumbnail />
+        </div>
+      </div>
 
-const TokenListSkeleton = ({round}: {round: boolean}) => (
```
caropinzonsilva, line 5: I think we needed this `round` prop in order to have a different loading state for locked vs unlocked. I didn't look a lot into it, but I wanted to point it out just in case!

caropinzonsilva (approved):
Looks good! I 🎩 in Storybook: https://639b1f308693132693d9b82c-rlsioyvoum.chromatic.com/?path=/story/tokengate-discount-unlocked--order-limit

Soleone:
Okay, perfect, thanks for clarifying, I was just double checking! 👍

